PR #3, opened 2014-07-03
Fix variable shadowing warning

MarcelRaad:
Microsoft Visual C++ 14 has a new warning:
"warning C4456: declaration of 'u' hides previous local declaration".


MarcelRaad:
Thank you!

PR #9, opened 2014-09-09
Feature: hyperexponential distribution

sguazt:
Hello,
I've recently added the hyperexponential distribution to Boost.Math/Statistical Distribution (with the precious feedback of John Maddock and Paul Bristow).
Here is the pull request log: https://github.com/boostorg/math/pull/7
The contribution has been already  accepted and currenlty it is in the `develop` branch (there may be some additional refinements, but the class is pretty stable).

I thought that it would be very nice to have in Boost the random variate generator counterpart.

So, here it is :)

If you like this idea, can you consider this pull request?

The development includes the random generator class + doxygen doc and two test units.
Note that I had to write my own test units in a similar fashion to what has be done for the `discrete_distribution`).

Thank you so much.

Best,
  Marco


swatanabe:
I think check_probabilities should be removed entirely.  The following asserts:

```c++
#include <boost/random/hyperexponential_distribution.hpp>

int main() {
    std::vector<double> weights(1023, 1.0);
    boost::random::hyperexponential_distribution<> dist(weights, weights);
}
```

swatanabe:
I'd strongly prefer to avoid external links outside the References section.  For links to other boost libraries, you can use boost:/libs/....  For UniformRandomNumberGenerator use \uniform_random_number_generator (This is defined in the doc Jamfile).

include/boost/random/hyperexponential_distribution.hpp
```diff
@@ -0,0 +1,797 @@
...
+            parm.probs_.swap(tmp);
+            hyperexp_detail::normalize(parm.probs_);
+
+            is >> std::ws;
+
+            tmp.clear();
+            detail::read_vector(is, tmp);
+            if (!is)
+            {
+                return is;
```
swatanabe: Reading probs_ but not rates_ will leave the object in an invalid state.

include/boost/random/hyperexponential_distribution.hpp
```diff
@@ -0,0 +1,797 @@
...
+            }
+            parm.rates_.swap(tmp);
+
+            //pre: vector size conformance
+            assert(parm.probs_.size() == parm.rates_.size());
```
swatanabe: assert is for programming errors.  assert should never be used to validate input.

sguazt:
Steven,

Thank you for your feedback.
I've tried to fix some of the issues you've highlighted, specifically:

- Fixed doc: removed (or moved to a local *References* section) external links, and replaced internal links as you suggested.
- Improved the function for reading a `param_type` from an input stream so that the internal state doesn't remain inconsistent (I've also added some tests for this in the `test_streaming` test case).
- Improved the `check_probabilities` function so that the check against sum of probabilities is some according to the relative magnitude of the two numbers to be checked, that is:
   `x` is approximately equal to `y` with respect to tolerance `tol`, if and only if `abs(x-y)/max(abs(x),abs(y)) <= tol`.
I've found this method in Knuth "The Art of Computer Programming, Vol 3", 3rd Ed, 1998 (Sec. 4.2.2).
 I've added a test case named `test_normalization`. On my machine, the test passes. But, I admit that the test also passed with the old version of `check_probabilities`. So, before celebrating, I wait for your feedback.

There is still an unresolved issue about the assertions for which I would ask you an opinion.
Do you suggest,

1. to replace assertions with exceptions, 
2. to completely remove them,
3. or what else

?

Thanks,

Marco

swatanabe:
AMDG

On 03/09/2015 10:40 AM, Marco Guazzone wrote:
> - Improved the function for reading a `param_type` from an input stream so that the internal state doesn't remain inconsistent (I've also added some tests for this in the `test_streaming` test case).

It's still possible for the internal state to
be inconsistent if is.exceptions has been set.

> - Improved the `check_probabilities` function so that the check against sum of probabilities is some according to the relative magnitude of the two numbers to be checked, that is:
>    `x` is approximately equal to `y` with respect to tolerance `tol`, if and only if `abs(x-y)/max(abs(x),abs(y)) <= tol`.
> I've found this method in Knuth "The Art of Computer Programming, Vol 3", 3rd Ed, 1998 (Sec. 4.2.2).
>  I've added a test case named `test_normalization`. On my machine, the test passes. But, I admit that the test also passed with the old version of `check_probabilities`. So, before celebrating, I wait for your feedback.
> 

The test that failed for me was something like:

std::vector<double> v(1001, 1.0);
boost::random::hyperexponential_distribution<> dist(v, v);

I think your new assert will still fire for this.
In fact, sinply increasing the size of the
vector is enough to make the assertion fire
for any constant value of tol.

> There is still an unresolved issue about the assertions for which I would ask you an opinion.
> Do you suggest,
> 
> 1. to replace assertions with exceptions, 
> 2. to completely remove them,
> 3. or what else
> 

You should set the fail bit of the stream.

In Christ,
Steven Watanabe

sguazt:
Hi Steven,
I've changed again  the function `operator>>` for reading the `param_type`. This function now operates in the following manner:

- If an I/O error occurs (possibly with exception throwing), the state of `parm` (i.e., the object to be read) is left unchanged, and specifically is the same one that `parm` has at the beginning of the function's execution.
- The state of `parm` is changed only after having completely read the whole input.

Better now?

Concerning, the check for probability normalization, even if the case you've suggested (i.e., `1001` and `1`) on my machine does not cause problem, I agree with you that if it fails on some architecture it is better to remove it.
So now, `check_probabilities` only checks if each probability is a finite number and is in the range of `[0,1]`.
Do you agree with these choices?

Best,

Marco

swatanabe:
AMDG

On 03/11/2015 03:52 PM, Marco Guazzone wrote:
> Hi Steven,
> I've changed again  the function `operator>>` for reading the `param_type`. This function now operates in the following manner:
> 
> - If an I/O error occurs (possibly with exception throwing), the state of `parm` (i.e., the object to be read) is left unchanged, and specifically is the same one that `parm` has at the beginning of the function's execution.
> - The state of `parm` is changed only after having completely read the whole input.
> 
> Better now?
> 

  That looks good.  My only remaining concern with
operator>> is how you fix up different sized
vectors.  I'm not saying that it's wrong, but
I'd like to hear the justification for why it's
the right behavior.  In particular:
- Why is this better than an error?
- Why fill the smaller range rather than
  truncating the larger range?
- Why fill with 1 rather than some other
  value (say 0)?

  If we agree that this kind of fixing up is
the right thing to do, then I'll need to
make the piecewise_constant and piecewise_linear
distributions match.

> Concerning, the check for probability normalization, even if the case you've suggested (i.e., `1001` and `1`) on my machine does not cause problem, I agree with you that if it fails on some architecture it is better to remove it.
> So now, `check_probabilities` only checks if each probability is a finite number and is in the range of `[0,1]`.
> Do you agree with these choices?
> 

I'm surprised that you can't duplicate this.  I'm not exactly
using an unusual architecture.  Here's the complete test program:

#include <boost/random/hyperexponential_distribution.hpp>
#include <vector>

int main() {
    std::vector<double> v(1001, 1.0);
    boost::random::hyperexponential_distribution<> d(v, v);
}

It fails with an assertion with MSVC 2013 and gcc 4.8.3.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 03/09/2015 10:40 AM, Marco Guazzone wrote:
> 
> - Fixed doc: removed (or moved to a local *References* section) external links, and replaced internal links as you suggested.
> 

I'd rather not refer to cppreference.com at all.  If you're
going to reference anything, the official source is:
C++11 (ISO/IEC 14882-2011)
8.5.4 [dcl.init.list] and
24.2.3 [input.iterators]

In Christ,
Steven Watanabe

sguazt:
Hi Steven,

> My only remaining concern with operator>> is how you fix up different sized vectors. I'm not saying
> that it's wrong, but I'd like to hear the justification for why it's the right behavior. In particular:
> - Why is this better than an error?

In Boost.Random, I saw very few places where exceptions are thrown (i.e., in `seed_impl.hpp` and `random_device.cpp`). So I thought that the policy in Boost.Random is to avoid throwing exception.
However, if we are allowed to throw exceptions, then I agree with you that it is a better choice than the current kind of *self-healing* behavior.
In this case, I would opt for `std::runtime_error`, would you?

> - Why fill the smaller range rather than truncating the larger range?

No particular reason. This is an arbitrary choice.

> - Why fill with 1 rather than some other value (say 0)?

Hmmm. Indeed, for probabilities `0` may be a good choice if we want that the corresponding exponential distribution will never be selected. On the other hand, to make sure the corresponding exponential distribution is selected at some point in the future, we need to set a value `> 0`. Again, my choice is totally arbitrary.
By the way, I've found that setting probabilities to `1` without subsequently calling the normalization function is a bug (thank you for pointing it out). So, whatever value we choose for probabilities (except for `0`), the correct way is to call the normalization function later (e.g., at line 517). Before fixing it, I see what we decide about throwing exceptions.
Concerning rates, `0` is not an admissible value (rates must be `> 0`). So `1` is just a matter of choice for setting rates to a valid value.

Best,

Marco

sguazt:
> I'm surprised that you can't duplicate this. I'm not exactly using an unusual architecture.

Don't know why the test runs fine on my machine and does not on yours.
To test this issue I wrote the test case `test_normalization` in `test_hyperexponential_distribution.cpp`.  Now, I've also added in the same test an explicit call to `check_params` (which in turn calls `check_probabilities`), in case `assert`s were not enabled by the compiler. As you can see, the test case checks against the two cases you've reported above.
On my machine it passed without problems.
My system is:

- Intel(R) Core(TM) i7-3540M
- Fedora Linux 21 x86_64
- GCC 4.9.2 20150212 (Red Hat 4.9.2-6)

Note (just to clarify), in the current commit, the test case and the offending check in `check_probabilities` are still commented. So, in the case you want to run such test case, you need to uncomment them:

- lines 242:261 in `test_hyperexponential_distribution.cpp`
- lines 95:100 in `hyperexponential_distribution.hpp`

Best,

Marco

sguazt:
> I'd rather not refer to cppreference.com at all. If you're going to reference anything, the official
> source is: C++11 (ISO/IEC 14882-2011) 8.5.4 [dcl.init.list] and 24.2.3 [input.iterators]

Sure, definitely better.
I've fixed it

Thank you

Marco

swatanabe:
AMDG

On 03/12/2015 05:35 AM, Marco Guazzone wrote:
>> My only remaining concern with operator>> is how you fix up different sized vectors. I'm not saying
>> that it's wrong, but I'd like to hear the justification for why it's the right behavior. In particular:
>> - Why is this better than an error?
> 
> In Boost.Random, I saw very few places where exceptions are thrown (i.e., in `seed_impl.hpp` and `random_device.cpp`). So I thought that the policy in Boost.Random is to avoid throwing exception.
> However, if we are allowed to throw exceptions, then I agree with you that it is a better choice than the current kind of *self-healing* behavior.
> In this case, I would opt for `std::runtime_error`, would you?
> 

The convention for stream operators is to use
is.setstate(std::ios_base::failbit);

This may throw an exception if the stream
has been configured to throw.

>> - Why fill the smaller range rather than truncating the larger range?
> 
> No particular reason. This is an arbitrary choice.
> 
>> - Why fill with 1 rather than some other value (say 0)?
> 
> Hmmm. Indeed, for probabilities `0` may be a good choice if we want that the corresponding exponential distribution will never be selected. On the other hand, to make sure the corresponding exponential distribution is selected at some point in the future, we need to set a value `> 0`. Again, my choice is totally arbitrary.
> By the way, I've found that setting probabilities to `1` without subsequently calling the normalization function is a bug (thank you for pointing it out). So, whatever value we choose for probabilities (except for `0`), the correct way is to call the normalization function later (e.g., at line 517). Before fixing it, I see what we decide about throwing exceptions.
> Concerning rates, `0` is not an admissible value (rates must be `> 0`). So `1` is just a matter of choice for setting rates to a valid value.
> 

Okay.  '0' also won't work for probabilities if the
given probability vector is empty.  If we can't use
'0', then '1' is probably the best choice.

In Christ,
Steven Watanabe

sguazt:
Hi,

Now, in case of nonconformant vectors' size, the `failbit` state is set.
Also, I've expanded the `test_streaming` test case to cover tests (1) when `exceptions(failbit)` is either disabled or enabled, and (2) to test the reading of both `param_type` and `hyperexponential_distribution` object.

Best,

Marco

sguazt:
Hi Steven,

Have you had time to check my last changes?
Any other change in order to finally get hyper-exponential merged into the main branch?

Thank you so much.

Best,

Marco

swatanabe:
AMDG

On 03/22/2015 02:50 AM, Marco Guazzone wrote:
>
> Have you had time to check my last changes?
> Any other change in order to finally get hyper-exponential merged into the main branch?
>

I'm getting errors with C++11:
test_hyperexponential_distribution.cpp(68) : error C3861: 'list_of': 
identifier not found

Also, on the iterator constructor:
// Note that we allow different argument types here to allow for
//  construction from an array plus a pointer into that array.
hyperexponential_distribution(RateIterT const& rate_first,
                               RateIterT2 const& rate_last);

This doesn't actually work, because you can end up with

RateIterT = double[N]
RateIterT2 = double *

rate_first then decays to const double *, which
makes the call to std::distance ambiguous.

In Christ,
Steven Watanabe

sguazt:
Thank you Steven.

> I'm getting errors with C++11: test_hyperexponential_distribution.cpp(68) : error C3861: 'list_of': identifier not found 

Indeed, some of the `list_of` occurrences were without `boost::assign` qualifier.
I've corrected the code, but I was not able to test it. How do you tell `bjam` to compile with C++11?

> This doesn't actually work, because you can end up with
> RateIterT = double[N]
> RateIterT2 = double * rate_first
> then decays to const double *, which makes the call to std::distance ambiguous.

Yes, you're right. I've fixed it by removing the `const&` and the second iterator type `RateIterT2`.
Also, I've added a related test case in `test_hyperexponential_distribution.cpp` (see lines `99` to `106`).

Cheers,

Marco

swatanabe:
AMDG

On 03/29/2015 07:28 AM, Marco Guazzone wrote:
> Thank you Steven.
> 
>> I'm getting errors with C++11: test_hyperexponential_distribution.cpp(68) : error C3861: 'list_of': identifier not found 
> 
> Indeed, some of the `list_of` occurrences were without `boost::assign` qualifier.
> I've corrected the code, but I was not able to test it. How do you tell `bjam` to compile with C++11?
> 

With gcc, cxxflags=-std=c++11

In Christ,
Steven Watanabe

sguazt:
Thank you.

I've fixed another bug in a test case.
Now all tests compile even with C++11.

Cheers,

Marco

swatanabe:
AMDG

The only remaining problems are with Doxygen:
- parm is #defined to param by the doxyfile,
  which results in messages like
  boost/random/hyperexponential_distribution.hpp:688: warning: argument
'parm' of command @param is not found in the argument list of
boost::random::hyperexponential_distribution< RealT
>::hyperexponential_distribution(param_type const &param)

- Doxygen is generating <ndash/> which is not correctly
  translated by BoostBook/DocBook.  I'll fix this one in
  doxygen2boostbook.xsl.

In Christ,
Steven Watanabe

sguazt:
Hi Steven,

I think to have fixed the above issues:

1. I've renamed `parm` into `param`
2. I've changed `--` into `-` in bibliographic references

Now, `bjam doc` stops to complain on my machine.
Let me know if there are other problems

Cheers,

Marco

sguazt:
Thank you, Steven.

Just to know... You closed this pull request without merging.
Do you plan to merge soon, or do you think this is not ready to be merged yet?

Cheers,

Marco

swatanabe:
AMDG

On 04/03/2015 01:39 AM, Marco Guazzone wrote:
> 
> Just to know... You closed this pull request without merging.
> Do you plan to merge soon, or do you think this is not ready to be merged yet?
> 

The changes are in develop.  I squashed them.

In Christ,
Steven Watanabe

sguazt:
Thanks!

Marco

PR #10, opened 2014-11-18
Fixed ticket #10110

maximiliank:
Fixed https://svn.boost.org/trac/boost/ticket/10110
Negative binomial distribution works now for real positive parameters k.
Added assertions on the parameters, i.e. k > 0 and 0 < p < 1.


swatanabe:
This is a breaking change and makes it incompatible with `std::negative_binomial_distribution`.

PR #13, opened 2015-01-25
Add includes for mpl types - replying on implicit includes via type trai...

jzmaddock:
...ts won't work going forward.


swatanabe:
Fixed in develop and master.

PR #16, opened 2015-03-08
Multiprecision support in Boost.Random

jzmaddock:
Steven, this is I hope the completed pull request.  Tested with MSVC 10, 11, 12, GCC-4.8, 4.9 in 03 and 11 modes.  Intel-15 (win and linux), clang/linux, and oracle/solaris.

Thankfully very few changes were required for the floating-point distros.  I also got discrete_distribution working with both large ints, and multiprecision float weights (though I'm not sure what the point would be).

And finally, a few headers were not being included by boost/random.hpp which is also fixed by this.

### DIFF
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -217,7 +217,7 @@ void generate_impl(Engine& eng, Iter first, Iter last, boost::mpl::false_)
 template<class Engine, class Iter>
 void generate(Engine& eng, Iter first, Iter last)
 {
-   return detail::generate_impl(eng, first, last, boost::random::traits::is_integral<typename Engine::result_type>());
+    return detail::generate_impl(eng, first, last, boost::random::traits::is_integral<typename Engine::result_type>());
 }
 
 
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -313,7 +313,7 @@ void fill_array_int_impl(Iter& first, Iter last, UIntType (&x)[n])
 template<int w, std::size_t n, class Iter, class IntType>
 inline void fill_array_int_impl(Iter& first, Iter last, IntType (&x)[n], boost::mpl::true_)
 {
-   typedef typename boost::boost::random::traits::make_unsigned<IntType>::type unsigned_array[n];
+    typedef typename boost::random::traits::make_unsigned<IntType>::type unsigned_array[n];
     fill_array_int_impl<w>(first, last, reinterpret_cast<unsigned_array&>(x));
 }
 
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -279,7 +279,8 @@ void seed_array_int_impl(SeedSeq& seq, UIntType (&x)[n])
 template<int w, std::size_t n, class SeedSeq, class IntType>
 inline void seed_array_int_impl(SeedSeq& seq, IntType (&x)[n], boost::mpl::true_)
 {
-    typedef typename boost::random::traits::make_unsigned<IntType>::type unsigned_array[n];
+    BOOST_STATIC_ASSERT_MSG(boost::is_integral<IntType>::value, "Sorry but this routine has not been ported to non built-in integers as it relies on a reinterpret_cast.");
+    typedef typename boost::make_unsigned<IntType>::type unsigned_array[n];
     seed_array_int_impl<w>(seq, reinterpret_cast<unsigned_array&>(x));
 }
 
@@ -313,7 +314,8 @@ void fill_array_int_impl(Iter& first, Iter last, UIntType (&x)[n])
 template<int w, std::size_t n, class Iter, class IntType>
 inline void fill_array_int_impl(Iter& first, Iter last, IntType (&x)[n], boost::mpl::true_)
 {
-    typedef typename boost::random::traits::make_unsigned<IntType>::type unsigned_array[n];
+    BOOST_STATIC_ASSERT_MSG(boost::is_integral<IntType>::value, "Sorry but this routine has not been ported to non built-in integers as it relies on a reinterpret_cast.");
+    typedef typename boost::make_unsigned<IntType>::type unsigned_array[n];
     fill_array_int_impl<w>(first, last, reinterpret_cast<unsigned_array&>(x));
 }
 
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -326,7 +326,7 @@ inline void fill_array_int_impl(Iter& first, Iter last, IntType (&x)[n], boost::
 template<int w, std::size_t n, class Iter, class IntType>
 inline void fill_array_int(Iter& first, Iter last, IntType (&x)[n])
 {
-    fill_array_int_impl<w>(first, last, x, boost::is_signed_helper<IntType>());
+    fill_array_int_impl<w>(first, last, x, boost::random::traits::is_signed<IntType>());
 }
 
 template<int w, std::size_t n, class RealType>
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -130,12 +130,28 @@ void generate_from_real(Engine& eng, Iter begin, Iter end)
         }
     }
 }
+//
+// Helper trait which handles unsigned multiprecision types
+// as well as builtin ones:
+//
+template <class T, bool intrinsic>
+struct make_unsigned_number
+{
+   typedef typename boost::make_unsigned<T>::type type;
+};
+template <class T>
+struct make_unsigned_number<T, false>
+{
+   BOOST_STATIC_ASSERT(std::numeric_limits<T>::is_specialized);
+   BOOST_STATIC_ASSERT(std::numeric_limits<T>::is_signed == false);
+   typedef T type;
+};
 
 template<class Engine, class Iter>
 void generate_from_int(Engine& eng, Iter begin, Iter end)
 {
     typedef typename Engine::result_type IntType;
-    typedef typename boost::make_unsigned<IntType>::type unsigned_type;
+    typedef typename make_unsigned_number<IntType, boost::is_integral<IntType>::value>::type unsigned_type;
     int remaining_bits = 0;
     boost::uint_least32_t saved_bits = 0;
     unsigned_type range = boost::random::detail::subtract<IntType>()((eng.max)(), (eng.min)());
--- a/include/boost/random/detail/seed_impl.hpp
+++ b/include/boost/random/detail/seed_impl.hpp
@@ -19,9 +19,7 @@
 #include <boost/config/no_tr1/cmath.hpp>
 #include <boost/integer/integer_mask.hpp>
 #include <boost/integer/static_log2.hpp>
-#include <boost/type_traits/is_signed.hpp>
-#include <boost/type_traits/is_integral.hpp>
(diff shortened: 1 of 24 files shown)


jzmaddock:
Forgot to mention, you need multiprecision develop branch to build against this.  I also need to make changes to the multiprecision lib tests/headers once this goes in.

swatanabe:
AMDG

I'm getting warnings from msvc because random/detail/traits.hpp
has #include enable_warnings.hpp without a corresponding
disable_warnings.hpp

In Christ,
Steven Watanabe

jzmaddock:
> I'm getting warnings from msvc because random/detail/traits.hpp
> has #include enable_warnings.hpp without a corresponding
> disable_warnings.hpp

Oops, pushed the obvious fix.

John.

swatanabe:
AMDG

On 03/08/2015 12:54 PM, jzmaddock wrote:
> Steven, this is I hope the completed pull request.  Tested with MSVC 10, 11, 12, GCC-4.8, 4.9 in 03 and 11 modes.  Intel-15 (win and linux), clang/linux, and oracle/solaris.
> 

MSVC 9 and GCC 4.6 are also okay.

In Christ,
Steven Watanabe

include/boost/random/detail/seed_impl.hpp
```diff
@@ -219,7 +217,7 @@ void generate_impl(Engine& eng, Iter first, Iter last, boost::mpl::false_)
 template<class Engine, class Iter>
 void generate(Engine& eng, Iter first, Iter last)
 {
-    return detail::generate_impl(eng, first, last, boost::is_integral<typename Engine::result_type>());
+   return detail::generate_impl(eng, first, last, boost::random::traits::is_integral<typename Engine::result_type>());
```
swatanabe: I'd prefer not to change the indentation here

include/boost/random/detail/seed_impl.hpp
```diff
@@ -281,7 +279,7 @@ void seed_array_int_impl(SeedSeq& seq, UIntType (&x)[n])
 template<int w, std::size_t n, class SeedSeq, class IntType>
 inline void seed_array_int_impl(SeedSeq& seq, IntType (&x)[n], boost::mpl::true_)
 {
-    typedef typename boost::make_unsigned<IntType>::type unsigned_array[n];
+    typedef typename boost::random::traits::make_unsigned<IntType>::type unsigned_array[n];
     seed_array_int_impl<w>(seq, reinterpret_cast<unsigned_array&>(x));
```
swatanabe: We can get away with reinterpret_cast for builtin signed/unsigned types.  Not so much for user-defined types.

include/boost/random/detail/traits.hpp
```diff
@@ -0,0 +1,107 @@
+/* boost random/detail/seed.hpp header file
```
swatanabe: Should be traits.hpp.  Also, is this a public header or not?  If it's not public, it should not be included in the documentation and should not be visible to users.  If it is public, it shouldn't be under detail/.

include/boost/random/detail/uniform_int_float.hpp
```diff
@@ -60,7 +60,7 @@ class uniform_int_float
     result_type operator()()
     {
         base_result range = static_cast<base_result>((max)())+1;
-        return static_cast<result_type>(_rng() * range);
+        return static_cast<result_type>(static_cast<base_result>(_rng() * range));
```
swatanabe: I'm somewhat curious why the expression templates don't support these conversions, when multiprecision::number itself does.  I do notice that detail::expression has an (explicit) conversion to bool.

include/boost/random/detail/vector_io.hpp
```diff
@@ -29,6 +29,7 @@ void print_vector(std::basic_ostream<CharT, Traits>& os,
         iter = vec.begin(),
         end =  vec.end();
     os << os.widen('[');
+    os << os.widen(' ');
```
swatanabe: Is it absolutely necessary to change the format?  Have you at least checked that the original format is still readable by the new code.

test/Jamfile.v2
```diff
@@ -132,3 +132,9 @@ explicit statistic_tests.exe ;
 
 install statistic_tests : statistic_tests.exe : <install-type>EXE <location>. ;
 explicit statistic_tests ;
+
+#
+# Multiprecision tests:
+#
+run multiprecision_int_test.cpp /boost//unit_test_framework ;
+run multiprecision_float_test.cpp /boost//unit_test_framework : : : release ;
```
swatanabe: Why is this test run in release mode?

include/boost/random/detail/seed_impl.hpp
jzmaddock: Editor SNAFU, fix pushed.

include/boost/random/detail/seed_impl.hpp
jzmaddock: Over aggressive search and replace at fault: reverted that change and one other, and added static_asserts to validate preconditions.

include/boost/random/detail/traits.hpp
jzmaddock: > In include/boost/random/detail/traits.hpp 
> <https://github.com/boostorg/random/pull/16#discussion_r26088928>:
>
> > @@ -0,0 +1,107 @@
> > +/* boost random/detail/seed.hpp header file
>
> Should be traits.hpp. Also, is this a public header or not? If it's 
> not public, it should not be included in the documentation and should 
> not be visible to users. If it is public, it shouldn't be under detail/.
>
OK, moved to boost/random.  This is pretty much a "documented detail", I 
wouldn't expect anyone to need to specialize these traits, but you never 
know.  Better documented than not IMO.

include/boost/random/detail/uniform_int_float.hpp
jzmaddock: > I'm somewhat curious why the expression templates don't support these 
> conversions, when multiprecision::number itself does. I do notice that 
> detail::expression has an (explicit) conversion to bool.
>
Good question, it got overlooked I guess, conversions between number<> 
types are handled via converting constructors it's just conversions to 
other types (built in ones for example) that's missing.  I realise this 
would simply those horrible double-casts, let me experiment and see if I 
can add support for this to multiprecision without breaking anything.

test/Jamfile.v2
jzmaddock: > Why is this test run in release mode?
>

It's too darn slow in debug mode - those calls to seed() in the 
lagged_fibonacci sequences are the culprit - the lack of inlining really 
hurts in there (and they generate a *lot* of values).  Will add a comment.

include/boost/random/detail/vector_io.hpp
jzmaddock: > Is it absolutely necessary to change the format? Have you at least 
> checked that the original format is still readable by the new code.
>

I realise this is a horrible fix.  It's not going to be easy to fix 
upstream in multiprecision, but let me look at it again and see.

jzmaddock: >Is it absolutely necessary to change the format? Have you at least 
checked that the original format is still readable by the new code.

Reverted, and fixed upstream in multiprecision.

John.

include/boost/random/detail/uniform_int_float.hpp
jzmaddock: Fixed upstream in multiprecision and reverted this change: unfortunately adding explicit conversion operators to the expression templates caused new issues in chi_squared_distribution.hpp which required the "obvious" casts.  You win some you lose some I guess... whatever adding the explicit conversion operators to the expression templates is probably the right thing to do anyway.

I think that's all the points you raised addressed now?

swatanabe: AMDG

On 03/12/2015 12:00 PM, jzmaddock wrote:
>
> I think that's all the points you raised addressed now?
> 

I think so.  I'll try to review it again
by the end of this week.

In Christ,
Steven Watanabe

jzmaddock: > On 03/12/2015 12:00 PM, jzmaddock wrote: I think that's all the points 
> you raised addressed now?
> I think so. I'll try to review it again by the end of this week. In 
> Christ, Steven Watanabe

Thanks, no rush, as it's too late for this release anyway, John.

swatanabe: AMDG

On 03/12/2015 12:00 PM, jzmaddock wrote:
>> @@ -60,7 +60,7 @@ class uniform_int_float
>>      result_type operator()()
>>      {
>>          base_result range = static_cast<base_result>((max)())+1;
>> -        return static_cast<result_type>(_rng() * range);
>> +        return static_cast<result_type>(static_cast<base_result>(_rng() * range));
> 
> Fixed upstream in multiprecision and reverted this change: unfortunately adding explicit conversion operators to the expression templates caused new issues in chi_squared_distribution.hpp which required the "obvious" casts.  You win some you lose some I guess... whatever adding the explicit conversion operators to the expression templates is probably the right thing to do anyway.
> 

Did you push this update?  I don't see it on my end.

Also:
- There are several places that change to 3 space indent,
  when the surrounding code is all indented to 4 spaces.
- traits.hpp has a *DETAIL* #include guard.
- doc/Jamfile.v2 refers to detail/traits.hpp
- In independent_bits_engine, I'd prefer all the
  private functions to be at the end.  base_result_type
  should be hidden from doxygen (and I'd also prefer
  not to rename it).

In Christ,
Steven Watanabe

jzmaddock: > Did you push this update? I don't see it on my end.

My bad, now pushed.

 >There are several places that change to 3 space indent, when the 
surrounding code is all indented to 4 spaces.

Those should be fixed now.

 >traits.hpp has a *DETAIL* #include guard.
 >doc/Jamfile.v2 refers to detail/traits.hpp

Fixed.

 >In independent_bits_engine, I'd prefer all the private functions to be 
at the end.

Fixed.

 >base_result_type should be hidden from doxygen (and I'd also prefer 
not to rename it).

This one is trickier: base_result_type is used as the type used to seed 
by a single value: and since it just gets forwarded to the underlying 
generator, this is actually the correct type for a single valued seed.  
So the type *is* used in the interface, and is named to match the other 
public typedefs.  The alternative would be to make the seed type the 
same as the result type and explicitly cast to the base_result_type 
internally.  However, this has two key disadvantages - it's C++11 only, 
and it's also rather inefficient (we construct a wider/multiprecision 
type, just to narrow it again internally).  Oh, and it's illogical too ;)

BTW there are rather a lot of commits now, let me know if you want me to 
condense down to a single commit.

John.

swatanabe: AMDG

On 03/14/2015 05:55 AM, jzmaddock wrote:
>
>> base_result_type should be hidden from doxygen (and I'd also prefer 
> not to rename it).
> 
> This one is trickier: base_result_type is used as the type used to seed 
> by a single value: and since it just gets forwarded to the underlying 
> generator, this is actually the correct type for a single valued seed.  
> So the type *is* used in the interface, and is named to match the other 
> public typedefs.  The alternative would be to make the seed type the 
> same as the result type and explicitly cast to the base_result_type 
> internally.  However, this has two key disadvantages - it's C++11 only, 
> and it's also rather inefficient (we construct a wider/multiprecision 
> type, just to narrow it again internally).  Oh, and it's illogical too ;)
> 

Okay.  I don't like it, but there's no good solution.

> BTW there are rather a lot of commits now, let me know if you want me to 
> condense down to a single commit.
> 

That would be helpful.  A few commits should be
separate, though.  For example, adding missing
#includes to random.hpp.

In Christ,
Steven Watanabe

jzmaddock:
Closing - please see other "rationalised" PR.

PR #17, opened 2015-03-10
Port to Windows Runtime

john-peterson:
CryptGenRandom isn't in WinRT https://github.com/mirror/boost/issues/5#issuecomment-77984515

### DIFF
--- a/include/boost/random/detail/config.hpp
+++ b/include/boost/random/detail/config.hpp
@@ -16,13 +16,3 @@
     && !defined(BOOST_MSVC)
     #define BOOST_RANDOM_NO_STREAM_OPERATORS
 #endif
-
-// WinRT target.
-#if !defined(BOOST_RANDOM_WINDOWS_RUNTIME)
-# if defined(__cplusplus_winrt)
-#  include <winapifamily.h>
-#  if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP)
-#   define BOOST_RANDOM_WINDOWS_RUNTIME 1
-#  endif
-# endif
-#endif
--- a/include/boost/random/detail/config.hpp
+++ b/include/boost/random/detail/config.hpp
@@ -16,3 +16,13 @@
     && !defined(BOOST_MSVC)
     #define BOOST_RANDOM_NO_STREAM_OPERATORS
 #endif
+
+// WinRT target.
+#if !defined(BOOST_RANDOM_WINDOWS_RUNTIME)
+# if defined(__cplusplus_winrt)
+#  include <winapifamily.h>
+#  if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP)
+#   define BOOST_RANDOM_WINDOWS_RUNTIME 1
+#  endif
+# endif
+#endif
--- a/src/random_device.cpp
+++ b/src/random_device.cpp
@@ -26,10 +26,19 @@
 const bool boost::random::random_device::has_fixed_range;
 #endif
 
+// WinRT target.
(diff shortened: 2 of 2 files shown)


swatanabe:
AMDG

- There's no reason to use vector.  An array should
  work just fine, since the size is known at compile
  time
- Is there a reason to go through DataReader instead of
  just calling CryptographicBuffer::CopyToByteArray?
- I'd prefer not to use auto.  Boost.Random is still
  C++03 compatible.
- When was WINAPI_FAMILY_PARTITION defined?  I don't
  want to break compatibility with older compilers.
  I still test with VC9.
- Similarly, will it work with MinGW or clang-cl?
- Where are these components coming from?  I notice
  that you #ifdefed out the #include <windows.h>.
- Can these functions fail?  What happens if they do?
  The MSDN documentation is nearly useless.

In Christ,
Steven Watanabe

john-peterson:
>There's no reason to use vector. 
>Is there a reason to go through DataReader

i fixed it

>When was WINAPI_FAMILY_PARTITION defined? 

i fixed the config to the same as https://github.com/boostorg/asio/blob/master/include/boost/asio/detail/config.hpp

>I'd prefer not to use auto. 

there's no winrt compiler that doesn't support auto

>will it work with MinGW or clang-cl?

do they have an rt compiler?

>Where are these components coming from?

the namespaces comes from /ZW

>Can these functions fail? 

doesnt seem so

swatanabe:
AMDG

On 03/09/2015 11:20 PM, John Sebastian Peterson wrote:
>> I'd prefer not to use auto. 
> 
> there's no winrt compiler that doesn't support auto
> 

Okay.

>> will it work with MinGW or clang-cl?
> 
> do they have an rt compiler?
> 

My concern wasn't so much compiling for WinRT
as the use of an unknown macro breaking desktop
compilation.

In Christ,
Steven Watanabe

john-peterson:
the check for the macro __cplusplus_winrt should prevent that. it's only present when using /ZW

swatanabe:
AMDG

By array I meant unsigned char[sizeof(result)].
Also, I'd rather not put any kind of WinRT
configuration in the header when it's only
used by random_device.cpp.

In Christ,
Steven Watanabe

john-peterson:
done

>By array I meant unsigned char[sizeof(result)].

it doesn't accept that type

>https://msdn.microsoft.com/en-us/library/windows.security.cryptography.cryptographicbuffer.copytobytearray.aspx

>Type: array of Number [JavaScript] | System.Byte[] [.NET] | Platform::Array<uint8> [C++]

swatanabe:
AMDG

The most recent version of the WinRT #ifdefing
is broken:

compile-c-c++
..\..\..\bin.v2\libs\random\build\msvc-12.0\debug\threading-multi\random_device.obj
random_device.cpp
..\..\..\libs\random\src\random_device.cpp(117) : error C2653:
'CryptographicBuffer' : is not a class or namespace name

In Christ,
Steven Watanabe

john-peterson:
u need to add /ZW https://github.com/boostorg/build/pull/62

swatanabe:
AMDG

On 03/13/2015 04:03 PM, John Sebastian Peterson wrote:
> u need to add /ZW https://github.com/boostorg/build/pull/62
> 

I'm *NOT* building for WinRT.

In Christ,
Steven Watanabe

john-peterson:
k so winapifamily.h is included for all vc 2013 builds then https://github.com/jp9000/OBS/blob/master/extras/winapifamily.h#L34

swatanabe:
I've rebased this and applied it to develop.

PR #19, opened 2015-05-28
Avoid passing large parameters by value

tempoz:
Both the Engine and Distribution may be passed by const reference, which is best practice in this case to improve performance.
This was uncovered by Coverity and is related to defect CID72373.


swatanabe:
This change is invalid in C++03, as Engine may already be a reference type.  variate_generator is mostly obsolete anyway.

PR #20, opened 2015-08-31
Port to CloudABI.

EdSchouten:
CloudABI does not provide access to the global filesystem namespace. We
can simply call into arc4random_buf() to get high-quality random data.

### DIFF
--- a/src/random_device.cpp
+++ b/src/random_device.cpp
@@ -147,6 +147,29 @@ private:
   const std::string provider;
 };
 
+#elif defined(__CloudABI__)
+
+#include <stdlib.h>
+
+namespace {
+const char * const default_token = "";
+}
+
+// CloudABI does not provide access to the global filesystem namespace.
+// The only way to obtain random data is by calling arcrandom_buf().
+class boost::random::random_device::impl
+{
+public:
+  impl(const std::string & token) {}
+  ~impl() {}
+
+  unsigned int next() {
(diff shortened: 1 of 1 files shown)


src/random_device.cpp
```diff
@@ -147,6 +147,29 @@ class boost::random::random_device::impl
...
+#elif defined(__CloudABI__)
+
+#include <stdlib.h>
+
+namespace {
+const char * const default_token = "";
+}
+
+// CloudABI does not provide access to the global filesystem namespace.
+// The only way to obtain random data is by calling arcrandom_buf().
```
muggenhor: I think you meant "arc4random_buf" instead. I.e. with the 4 included.

jeking3:
Once #29 is merged, this could be added as a entropy provider concept to random device if we create a new file similar to https://github.com/boostorg/random/pull/29/files#diff-f2cd05667b88b5aa79ee2ec8de16f63f and then hook it into the automatic selection logic in random_device (include/boost/random/random_device.hpp, line 155).  

Would probably recommend we add CloudABI to the Boost.Predef library first as ``BOOST_PLAT_CLOUDABI`` for detection.

jeking3:
I submitted a pull request into Boost.Predef for CloudABI platform detection: https://github.com/boostorg/predef/pull/65

I can use that to make a very simple random device entropy provider that will work with #29.

PR #21, opened 2015-09-10
Add missing break statement.

mamoll:


mamoll:
See also https://svn.boost.org/trac/boost/ticket/11454

mamoll:
Fixed in 29e8bd59a24ac2f6023c3706916f829b0d416297

PR #22, opened 2015-12-16
No undesired auto linking

muggenhor:
Don't auto-link when the user requested that we don't

### DIFF
--- a/src/random_device.cpp
+++ b/src/random_device.cpp
@@ -47,9 +47,11 @@ using namespace Platform;
 using namespace Windows::Security::Cryptography;
 #endif
 
-#define BOOST_AUTO_LINK_NOMANGLE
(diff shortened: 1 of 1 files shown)


swatanabe:
This is definitely the wrong solution. The user doesn't (usually) control the build of the random library.

muggenhor:
FYI: this was a long long time ago. But what i remember is that this change was necessary for building and using this with WinCE. In that scenario we did control the build.

I don't recall the exact problem but WinCE had a lot of it's functions in different libraries from the desktop variant. So possibly Advapi32 didn't even exist there. It's use definitely caused build errors.

But it's a long time ago, and the project in question is in maintenance by completely different people now. So this is the only context I can provide from memory now

PR #24, opened 2016-06-03
Added constexpr support to min/max functions of random number generators. Fixes #12251

gblanco92:
This fixes https://svn.boost.org/trac/boost/ticket/12251

### DIFF
--- a/include/boost/random/additive_combine.hpp
+++ b/include/boost/random/additive_combine.hpp
@@ -59,12 +59,12 @@ public:
     /**
      * Returns the smallest value that the generator can produce
      */
-    static result_type min BOOST_PREVENT_MACRO_SUBSTITUTION ()
+    static BOOST_CONSTEXPR result_type min BOOST_PREVENT_MACRO_SUBSTITUTION ()
     { return 1; }
     /**
      * Returns the largest value that the generator can produce
      */
-    static result_type max BOOST_PREVENT_MACRO_SUBSTITUTION ()
+    static BOOST_CONSTEXPR result_type max BOOST_PREVENT_MACRO_SUBSTITUTION ()
     { return MLCG1::modulus-1; }
 
     /**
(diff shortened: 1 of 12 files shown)


mclow:
The code changes look good, but there's no tests!

gblanco92:
I will add tests for this!

mclow:
Ping?

rtimmons:
Is there any way to get this merged?

I've got a project that wants to use cross-platform `std::stuffle` and the lack of `constexpr min()` and `constexpr max()` on `boost::random::mt19937_64` and friends leads to compiler errors

```
.../usr/include/c++/v1/algorithm:2950:57: error:
  constexpr variable '_Rp' must be initialized by a constant expression
    static _LIBCPP_CONSTEXPR const _Working_result_type _Rp = _Engine::max() - _Engine::min()
```

I've verified that this change fixes the compiler error at least for my limited selection of platforms (latest gcc and clang on osx and a few linuxes).

mclow:
Still has no tests....

rtimmons:
Thanks @mclow - do you have any input on how this should be tested to satisfaction?

-   My instincts say it's enough to e.g. assert `additive_combine::min()` (and similar) can be evaluated in a constexpr context. 

    Is there a way to do this that respects the `BOOST_CONSTEXPR` macro? Is it sufficient to rely on the `_NO_*_CONSTEXPR` defs?

    ```c++
    #if !defined(BOOST_NO_CXX11_CONSTEXPR) && !defined(BOOST_NO_CXX14_CONSTEXPR)
    // test that xyz::min(),xyz::max() can be evaluated in constexpr context
    #endif
    ```

- Alternatively we could assert that `std::shuffle` or other STL algorithms work with these RNGs. But that may not be direct enough for this and/or depends on a particular STL requiring `min/max` to be `constexpr` (which [the reference](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) doesn't actually seem to require)

mclow:
> My instincts say it's enough to e.g. assert additive_combine::min() (and similar) can be evaluated in a constexpr context.

Can they be evaluated, and do they return a sane answer.
`static_assert` is probably your friend here.

PR #27, opened 2017-05-13
Conditionally replace deprecated/removed C++98 std::bind2nd by std::b…

DanielaE:
…ind, and std::random_shuffle by an adapted std::shuffle.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/random/uniform_on_sphere.hpp
+++ b/include/boost/random/uniform_on_sphere.hpp
@@ -225,8 +225,12 @@ public:
                     }
                 } while(sqsum == 0);
                 // for all i: result[i] /= sqrt(sqsum)
-                std::transform(_container.begin(), _container.end(), _container.begin(),
-                               std::bind2nd(std::multiplies<RealType>(), 1/sqrt(sqsum)));
+                RealType inverse_distance = 1 / sqrt(sqsum);
+                for(typename Cont::iterator it = _container.begin();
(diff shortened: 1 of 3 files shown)


jzmaddock:
> > M test/test_old_uniform_int_distribution.cpp (39)
> > M test/test_random_number_generator.cpp (38)
> >
>
> Adapting these to use std::shuffle makes no sense,
> because they are specifically testing compatibility
> with std::random_shuffle. If std::random_shuffle
> doesn't exist, there's no point to these tests and
> they can just be #ifdef'ed out entirely.
>

FYI we have |BOOST_NO_CXX98_RANDOM_SHUFFLE for just that purpose.

HTH, John.
|

---
This email has been checked for viruses by AVG.
http://www.avg.com

DanielaE:
Steve, I hope you like this more!
John, of course ...

swatanabe:
AMDG

On 05/13/2017 11:04 AM, Daniela Engert wrote:
> Steve, I hope you like this more!
> John, of course ...
> 

  The random_shuffle changes look good, but
in C++03, a local class cannot be a template
parameter.

In Christ,
Steven Watanabe

DanielaE:
Steve, thx for this info - never heard about that before. Then I am probably going to make Sean unhappy ...

PR #36, opened 2018-02-15
Feature: Quasi-random generators, round II

justinas-vd:
1. Removed typedefs from boost namespace.
2. Upped supported dimensions for both Sobol and Niederreited base 2 qrngs.
3. Default sequence count is now 64 bit.
4. Cleaned up the warnings in doc generation.


include/boost/random/faure.hpp
```diff
@@ -0,0 +1,342 @@
...
+  }
+  return ilog;
+}
+
+inline std::size_t integer_pow(std::size_t base, std::size_t e)
+{
+  // Reuse the implementation of exponentiation by squaring, but
+  // there is a twist that powm actually computes base^e % m,
+  // so for m we plug-in a fully set bitset, whose modulo will be no-op.
+  return multiprecision::powm(base, e, ~std::size_t());
```
NAThompson: Maybe my previous advice to use `powm` was incorrect, if you don't need to use the modulus operator. Do you think it'll compile down to a no-op? If it does, will it do the wrong thing on overflow? (I just feel like there should be some way for code reuse on this one in boost or the stdlib, maybe that's naive. . . )

include/boost/random/faure.hpp
```diff
@@ -0,0 +1,342 @@
...
+
+private:
+  packed_uint_t qs_base;
+  RealType inv_qs_base;
+
+  // here we cache precomputed data; note that binomial coefficients have
+  // to be recomputed iff the integer part of the logarithm of seq changes,
+  // which happens relatively rarely.
+  std::vector<packed_uint_t> coeff; // packed upper (!) triangular matrix
+  std::vector<RealType> ytemp;
```
NAThompson: Does @swatanabe have a convention for member variables to follow, like `m_` prefixes?

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,351 @@
...
+    // Compute Niedderreiter base 2 lattice
+    for (std::size_t dim = 0; dim != dimension; ++dim)
+    {
+      const unsigned short poly = qrng_tables::niederreiter_base2::polynomial(dim);
+      if (static_cast<std::size_t>(poly) >
+          static_cast<std::size_t>(std::numeric_limits<IntType>::max())) {
+        boost::throw_exception( std::range_error("niederreiter_base2: polynomial value outside the given IntType range") );
+      }
+
+      const unsigned degree = multiprecision::detail::find_msb(poly); // integer log2(poly)
```
NAThompson: `boost/multiprecision/integer.hpp` has `msb`, which returns `detail::find_msb`

include/boost/random/faure.hpp
justinas-vd: I think it's possible to reuse code there, but it will require a bit of effort.

`powm` could probably be implemented like this:

```C++
template <class I1, class I2, class I3>
typename enable_if_c<is_integral<I1>::value && is_unsigned<I2>::value && is_integral<I3>::value, I1>::type
   powm(const I1& a, I2 b, I3 c)
{
   typedef typename detail::double_integer<I1>::type double_type;

   I1 x(1), y(a);
   double_type result;

   while(b > 0)
   {
      if(b & 1)
      {
         multiply(result, x, y);
         x = integer_modulus(result, c);
      }
      multiply(result, y, y);
      y = integer_modulus(result, c);
      b >>= 1;
   }
   return integer_modulus(x, c);
}
```

(Note `return integer_modulus(x, c);` instead of plain `return x % c;`)

And if  integer_modulus would support mpl::false_() as the second argument, it could return just the first one. Which would make everything work.

include/boost/random/niederreiter_base2.hpp
justinas-vd: Done.

justinas-vd:
Mostly, I am a bit concerned that Sobol tables could be excessively large. sobol_data.hpp is now 1.7 Mb, which is a bit hefty for an include, no?

NAThompson:
Yes, that could be a problem. Is there an algorithm to rapidly generate the direction numbers of the Sobol sequence at runtime? This would be more like what is done in boost.math: If you request more precision from (say) `boost::math::constants::pi<Real>()` than is cached, then an algorithm computes it. Otherwise, it just returns the number stored in the include file.

The `boost.math` prime table has 10,000 elements, and too my knowledge no one complains about it. But you are correct that 1.7mb is pretty big and should be avoided if possible.

Could you run a `time g++ sobol_test.cpp` and see if the compile time is unbearable?

justinas-vd:
time g++ faure_validate.cpp -O3
real	0m2,911s
user	0m2,717s
sys	0m0,193s

time g++ niederreiter_base2_validade.cpp -O3
real	0m3,571s
user	0m3,365s
sys	0m0,205s

time g++ sobol_validate.cpp -O3
real	0m4,311s
user	0m4,135s
sys	0m0,175s

Doesn't seem that the compile time is unbearable, but Sobol test certainly compiles slower than the lightweight (in comparison) Faure.

swatanabe:
AMDG

On 02/15/2018 09:29 AM, Justinas V. Daugmaudis wrote:
> Mostly, I am a bit concerned that Sobol tables could be excessively large. sobol_data.hpp is now 1.7 Mb, which is a bit hefty for an include, no?
> 

  It's too much.  Please also consider that this
will bloat object files and executables.  If you
really need to have all this data static, then
at least make it separately compiled.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 02/15/2018 09:09 AM, Nick wrote:
> NAThompson commented on this pull request.
> 
> <snip>
> +inline std::size_t integer_pow(std::size_t base, std::size_t e)
> +{
> +  // Reuse the implementation of exponentiation by squaring, but
> +  // there is a twist that powm actually computes base^e % m,
> +  // so for m we plug-in a fully set bitset, whose modulo will be no-op.
> +  return multiprecision::powm(base, e, ~std::size_t());
> 
> Maybe my previous advice to use `powm` was incorrect, if you don't need to use the modulus operator. Do you think it'll compile down to a no-op? If it does, will it do the wrong thing on overflow?

Overflow is guaranteed to wrap modulo 2^k for unsigned types.

> (I just feel like there should be some way for code reuse on this one in boost or the stdlib, maybe that's naive. . . )
> 

  This has got to be wrong.  The correct modulus is
std::numeric_limits<size_t>::max() + 1, which is
not representable in a size_t.  Making the modulus
too small by one, will cause the result to be 0
whenever it should be 0xFFFFFFFFFFFFFFFF.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 02/15/2018 08:21 AM, Justinas V. Daugmaudis wrote:
> 1. Removed typedefs from boost namespace.
> 3. Default sequence count is now 64 bit.

Upon further reflection, I think the correct way is:

template<class UIntType, int w>
class sobol_engine;

typedef sobol_engine<boost::uint_least32_t, 32> sobol32;
typedef sobol_engine<boost::uint_least64_t, 64> sobol64;

In Christ,
Steven Watanabe

justinas-vd:
Looks nice to me. But what's the purpose of template w parameter? How would it be used?

swatanabe:
AMDG

On 02/15/2018 11:41 AM, Justinas V. Daugmaudis wrote:
> Looks nice to me. But what's the purpose of template w parameter? How would it be used?
> 

  It indicates the bit width.  This is the convention
that the standard library uses, because the standard
does not guarantee that there are integer types of
any specific width.  It only guarantees minimum sizes.
Thus, uint_least32_t is always defined, but uint32_t
might not be.

In Christ,
Steven Watanabe

justinas-vd:
Okay, then for niederreiter_base2 and sobol this <int w> param should be used, then.
It will be used for the Gray code bits and for the counter (seq_count).

Thus we will have:
typedef sobol_engine<32> sobol32;
typedef sobol_engine<64> sobol64;

typedef niederreiter_base2_engine<32> niederreiter_base2_32;
typedef niederreiter_base2_engine<64> niederreiter_base2_64;

Then faure_engine needs to be parametrized as RealType and int w.
typedef faure_engine<float, 32> faure32;
typedef faure_engine<double, 64> faure64;

Something like this. What do you think?

justinas-vd:
@NAThompson Generating Sobol data for requested dimension is an absolute no. It is simply an overkill. Especially considering that we would have to maintain and update the solution with respect to the newest research in that area.

The best that I can think of is something like this:
1. you have some awesome table you want to have;
2. you pass this table to the generator.
Otherwise, the default reasonably-sized tables will be used.

I think that for niederreiter_base2 4720 dimensions is more than people have seen in a typical implementation. It's still a table of just 4720 unsigned shorts and thus bearable.

For Sobol I think I will downgrade to some dimension where numbers fit to unsigned short as well. Most likely I will simply cut off the newest 21201 dim tables to some reasonable size.

For aficionados and people pushing to the extremes I will simply provide a constructor argument to pass the table, and it will be their own responsibility from there. Sounds ok?

swatanabe:
AMDG

On 02/15/2018 12:28 PM, Justinas V. Daugmaudis wrote:
> <snip>
> Then faure_engine needs to be parametrized as RealType and int w.
> typedef faure_engine<float, 32> faure32;
> typedef faure_engine<double, 64> faure64;
> 

  That's not quite correct, as float and double only
have 24 and 53 bits in the mantissa, respectively.
It looks to me like faure is fundamentally floating
point rather than being an integer algorithm that's
been written using floating point numbers (like
lagged_fibonacci_01 or ranlux_01).  Thus, I don't
think the bit width actually makes sense.  w for
floating point generators only makes sense if you
could easily rewrite the algorithm to generate
the same sequence * 2^w as integers.

> Something like this. What do you think?
> 

In Christ,
Steven Watanabe

justinas-vd:
Sobol tables are down to 230Kb, yet they still support respectable 3667 dimensions. I think it's a nice balance.

include/boost/random/faure.hpp
justinas-vd: I will restore my previous code here, but let's keep in minds that with the suggestion above code reuse would be possible.

justinas-vd:
@NAThompson Now you can use whatever tables you want. I also made the tables part of the type definition, since otherwise generator comparison would not make sense.

doc/concepts.qbk
```diff
@@ -203,29 +203,29 @@ the generator, for example to re-run a test suite at a later time.]
...
 part of the square would be proportional to the number of points in the whole square.
-Such sequences share some properties of  random variables and in certain applications 
+Such sequences share some properties of  random variables and in certain applications
 such as the quasi-Monte Carlo method  their lower discrepancy is an important advantage.
 
-[note Quasi-random sequences are known to give efficient numerical integration 
-rules in many Bayesian statistical problems where the posterior distribution can be 
+[note Quasi-random sequences are known to give efficient numerical integration
+rules in many Bayesian statistical problems where the posterior distribution can be
 transformed into periodic functions on a hypercube.]
```
NAThompson: I do not believe there is any requirement that the function must be periodic.

NAThompson:
Cool, do you have an account on codereview.stackexchange.com? If not, do you mind if I communicate this code there on your behalf?

justinas-vd:
Nope, I don't have an account, and nope, I don't mind if anybody else reviews the code. The more eyes, the merrier.

doc/concepts.qbk
justinas-vd: I didn't imply that periodicity was a requirement. I just meant that QMC numerical integration is notably efficient there. But I think that we can broaden the description and make it less technical. 

For example:
Quasi-Monte Carlo method uses a low-discrepancy sequence such as the Niederreiter base 2 sequence, the Sobol sequence, or the Faure sequence among the others. The advantage of using low-discrepancy sequences is a probabilistically faster rate of convergence. Quasi-Monte Carlo has a rate of convergence around to O(1/N), whereas the rate for the Monte Carlo method is O(N^−(1/2)).

NAThompson:
Ok, here's the link:

[https://codereview.stackexchange.com/questions/187712/quasi-random-number-generators](https://codereview.stackexchange.com/questions/187712/quasi-random-number-generators)

Hopefully we get some good feedback; it is often the case that people give very nice feedback there, but this is a technically sophisticated effort and might be too much effort for the members of that forum.

doc/concepts.qbk
NAThompson: Oh, I didn't know that there was a better convergence rate for periodic integrands using qmc. If you do add this to the docs, make sure to get the logarithmic factors on the convergence: O(log(N)^d/N).

justinas-vd:
Seems good.

BTW, what did you think about the solution for tables? I hope we reached some middle ground between general usefulness and hardcore high-dimensionality.

NAThompson:
I think the solution for the tables is good. Could you link the 21000 dimensional tables in the docs?

Also, try to link to wayback machine, these academic web pages disappear all the time.

doc/concepts.qbk
justinas-vd: Nah, rewrote the note like this

[note Quasi-Monte Carlo method uses a low-discrepancy sequence such as the 
Niederreiter base 2 sequence, the Sobol sequence, or the Faure sequence among the others. 
The advantage of using low-discrepancy sequences is a probabilistically faster rate of 
convergence. Quasi-Monte Carlo has a rate of convergence close to '''<code>O(1/N)</code>''', 
whereas the rate of convergence for the Monte Carlo method, which uses a pseudo-random sequence, 
is '''<code>O(N<superscript>-0.5</superscript>)</code>'''.] 

I think "close to" is enough here.

justinas-vd:
Done.

NAThompson:
Ok, so just to regroup: 

- We've verified that this code is fast
- We've verified is has no memory corruptions or undefined behavior
- We've verified that this is better than existing open source alternatives, due to features and much higher dimensionality
- We have a use case which will consume this feature as soon as it is merged

But:

- The docs do not show an example. I don't think anyone goes to boost docs to learn about qrngs, they'll migrate from wikipedia and want to see a code example immediately. (Note to self: Need to edit wikipedia to link to these once they drop.)
- We do not have a unit test showing that the L2 discrepancy of the sequences satisfy known bounds. My own opinion is that this is a stronger test than matching published data, but this is not a showstopper for me.
- We still have a too-small prime table, and the prime table in boost.math doesn't seem to offer you the features you need. I'll talk to John Maddock about merging a sieve of Eratosthenes into boost.integer, which will hopefully solve this problem.
- There is no discussion on how to apply randomization to the sequences. I think there should be either a recommendation on how to do this in the docs, or a very simple Cranley-Patterson rotation should be supplied in the code. I have a Cranley-Patterson rotation [here](https://github.com/boostorg/math/blob/9d4752bc5671fad664f311f23fccbf49ec509b45/include/boost/math/tools/cranley_patterson_rotation.hpp) but I don't think it supports the typical idioms required in boost.random.

These are small problems. I still think that these will be the best feature landing in 1.67.

justinas-vd:
Thank you for your nice summary.

I just wanted to add a few words about discrepancy testing. You see, I have implemented an algorithm to compute upper bounds for the star discrepancy of an arbitrary set of n points in the s-dimensional unit cube as proposed in Eric Thiemard's paper: "Computing bounds for the star discrepancy", Computing  65 (2000) 2, 169-186.

I just didn't add this to this PR for several reasons:

1. This PR is big enough as it is. The d_star code itself is complex (~20 kbs of just dense code to review).
2. Furthermore, this raises the question: does discrepancy measuring work?
3. This leads to a convoluted scheme: a) verify discrepancy tester itself (with, say, a known Halton sequence of known bounds); b) then use the verifier to verify, say, niederreiter_base2. It should get the expected discrepancy of D_n^*(x) = O( (log n)^s∕n ) within some reasonable given epsilon.
4. Then we need to select a trio of parameters: number of points N, dimension s and epsilon eps for discrepancy tests. And this is not as straightforward as it seems, _especially_ if you want to run tests as part of automated test suite. You see, without careful fiddling those tests would be – more likely than not – exceedingly slow, but then this "careful fiddling" is fishy in the end. 

This is my rationale for not submitting that code here. Matching published data is a strong enough guarantee. It's fast, too.

NAThompson:
That sounds reasonable to me! It appears @swatanabe is busy with the Yap review, but we can hope he has time to merge before 1.67 closes.

swatanabe:
AMDG

On 02/15/2018 08:21 AM, Justinas V. Daugmaudis wrote:
> 
>     A test/niederreiter_base2_validade.cpp (372)

The file name is misspelled.

In Christ,
Steven Watanabe

swatanabe:
AMDG

  This PR needs to be rebased against develop.
multiprecision_float_test currently fails because
of changes to Boost.Test.

In Christ,
Steven Watanabe

justinas-vd:
Will do. File rename and PR reopen on develop branch.

PR #37, opened 2018-02-18
Feature: Quasi-random generators, round III

justinas-vd:
This is a rebranched (from master to develop) PR of QRNGs.


justinas-vd:
I have checked and rechecked, and I am fairly confident that the code should not eat anyone's dog.

By the way, @swatanabe , have you had a chance to see what the latest MS compiler says?

swatanabe:
It was fine with msvc9 and up last time I checked.  I haven't run the latest commit.

justinas-vd:
Should be fine, too, then. The latest commit simply allows the compiler to reuse div result, since this opcode yields both the quotient and the remainder of the division (in eax:edx registers). Apparently, x / y followed by x % y is a trivial optimization that gcc performs even at -O1.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+#ifdef BOOST_RANDOM_DOXYGEN
+  //=========================Doxygen needs this!==============================
+
+  //!Returns: The dimension of of the quasi-random domain.
+  //!
+  //!Throws: nothing.
+  std::size_t dimension() const { return base_t::dimension(); }
+
+  //!Requirements: *this is mutable.
```
swatanabe: That's a pointless thing to say, since the non-const signature says as much.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+  //!\code
+  //!X u = v;
+  //!for(int i = 0; i < N; ++i)
+  //!    u();
+  //!v.discard(N);
+  //!assert(u() == v());
+  //!\endcode
+  //!
+  //!Throws: overflow_error.
+  void discard(IntType z)
```
swatanabe: IntType is wrong.  It should always be unsigned long long (boost::uinmax_t).  Consider the case where IntType = uint8_t, with 2 dimensions.  The generator can produce 256 vectors, which is 512 elements, which is more than IntType can represent.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+}
+
+} // namespace nb2
+
+template<typename IntType, typename Nb2Table>
+struct niederreiter_base2_lattice
+{
+  typedef IntType value_type;
+
+  BOOST_STATIC_CONSTANT(unsigned, bit_count = std::numeric_limits<value_type>::digits);
```
swatanabe: This should be a template parameter, rather than being deduced from the size of the argument.  The goal is to make it so that the output of the generator is platform independent.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+  // We must have 0 <= Kj < E*J = M.
+  // The limit condition on Kj does not seem very relevant
+  // in this program.
+  size_type kj = pb_degree;
+
+  pb_degree = bitset_log2(pb);
+
+  // Now choose values of V in accordance with
+  // the conditions in section 3.3.
+  std::fill(v.begin(), v.begin() + kj, T());
```
swatanabe: There's also std::fill_n.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+  // Quoting from BFN: "Our program currently sets each K_q
+  // equal to eq. This has the effect of setting all unrestricted
+  // values of v to 1."
+  // Actually, it sets them to the arbitrary chosen value.
+  // Whatever.
+  for (size_type r = kj; r < pb_degree; ++r)
+    v[r] = arbitrary_element;
```
swatanabe: std::fill.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+  // Calculate the remaining V's using the recursion of section 2.3,
+  // remembering that the B's have the opposite sign.
+  for (size_type r = pb_degree; r < v.size(); ++r)
+  {
+    T term = T /*zero*/ ();
+    boost::dynamic_bitset<Block, Allocator> pb_c = pb;
+    for (int k = -pb_degree; k < 0; ++k, pb_c >>= 1)
+    {
+      if( pb_c.test(0) )
```
swatanabe: This is equivalent to pb.test(k + pb_degree).

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+        term ^= v[r + k];
+    }
+    v[r] = term;
```
swatanabe: The body of this loop is equivalent to:
```
dynamic_bitset<> tmp = (v >> (r - pb_degree));
tmp.resize(pb.size());
v[r] = (tmp & pb).count() % 2;
```
if you change v to be another dynamic_bitset.

include/boost/random/niederreiter_base2.hpp
justinas-vd: Yay, an excellent idea. How did I miss this.

include/boost/random/niederreiter_base2.hpp
justinas-vd: Changed v to dynamic_bitset, but kept the loop to check pb.test(k + pb_degree). Less temporaries.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+  typedef boost::dynamic_bitset<Block, Allocator> bitset_t;
+  typedef typename bitset_t::size_type size_type;
+
+  if (v.none())
+    throw std::invalid_argument("bitset_log2");
+
+  size_type up  = v.size() - 1;
+  size_type low = v.find_next(0);
+
+  // Binary lookup for the most significant set bit
```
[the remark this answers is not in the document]
justinas-vd: Alas, this function is necessary, since pb_degree does not increase monotonically. It is quite possible for kj (the previous pb_degree value) to be bigger than the next computed pb_degree. Try and printout the cases, when kj > pb_degree.

As for the reverse scan, you are right. I somehow didn't think of it. Done.

Nevertheless, I will note that this binary search technique is not totally without merit. It effectively cuts off all the lower bits if any upper bits are already set.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+        // Advance Niederreiter's state variables.
+        for (unsigned u = 0; u != degree && j != bit_count; ++u, ++j)
+        {
+          // Now C is obtained from V.  Niederreiter
+          // obtains A from V (page 65, near the bottom), and then gets
+          // C from A (page 56, equation (7)).  However this can be done
+          // in one step.  Here CI(J,R) corresponds to
+          // Niederreiter's C(I,J,R).
+          for (unsigned r = 0; r != bit_count; ++r) {
+            ci[r][j] = v[r + u];
```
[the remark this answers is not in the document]
justinas-vd: I get your idea that you want to put, say ci's column to the function that computes and, perhaps, avoid v altogether, but I don't think that idea will fly. At least not without some major algorithmic complications.

As for the iteration order, I iterate in column major here so that I could iterate in row major a few lines below. ;-) If I change the order in one place, I will need to change the order in another place.

include/boost/random/niederreiter_base2.hpp
```diff
@@ -0,0 +1,371 @@
...
+  {
+    detail::dimension_assert("Niederreiter base 2", dimension, Nb2Table::max_dimension);
+
+    // Initialize the bit array
+    bits.resize(boost::extents[bit_count][dimension]);
+
+    // Reserve temporary space for lattice computation
+    boost::multi_array<value_type, 2> ci(boost::extents[bit_count][bit_count]);
+
+    std::vector<value_type> v;
```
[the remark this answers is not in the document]
justinas-vd: Well noted.

include/boost/random/niederreiter_base2.hpp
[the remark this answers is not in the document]
justinas-vd: I have tested the performance of the revscan approach. I have to say that your reverse scan idea works slightly faster than bin search. Both revscan and bin search beat silly 

```C++
  size_type low = 0;
  size_type p = 0;

  // Linear lookup for the most significant set bit
  while (p != DynamicBitset::npos)
  {
    low = p;
    p = v.find_next(low);
  }

  return low;
```

hands down.

Timings on my machine for 10000000 random bitsets are like this:
bin_search: 2.35857s
rev_scan: 2.33807s
lin_search: 4.01772s

justinas-vd:
All done.

I also contemplated for some time to change 
```C++
boost::multi_array<bool, 2> ci(boost::extents[bit_count][bit_count]);
```
in niederreiter base2 lattice to, e.g.,
```C++
std::vector<bitset_type> ci(bit_count, bitset_type(bit_count));
```

From memory usage perspective assuming bit_count=64 the first case would use 4 Kbs of contiguous memory, and the second one would use 64*(32 + 8), where 32 bytes for sizeof(dynamic_bitset) and 8 more bytes for 64 bit data in the bitset, which yields 64*40 and this is sligtly smaller than 64*64, but the second version is not contiguous, not cache friendly and more expensive to allocate and deallocate.

justinas-vd:
@swatanabe , I think that niederreiter code is pretty compact now. I reversed ci bits, but now I can recover bitset values via to_ulong fn.

justinas-vd:
All the raised issues were addressed, please review again.

NAThompson:
I wonder if we can trade @swatanabe some bug fix triage for this getting merged; maintainer is a thankless job: They get to fix a bunch of code written by other people. But if we could credibly commit to fixing tickets, then perhaps it'll go a long way towards getting this into 1.67.

justinas-vd:
I see no problems to fix QRNGs should we get any tickets. But I have to say that Steven's review was invaluable. The code got leaner, more concise and faster.

So I don't mind waiting until he gets some time for code review. Bug fixing promises aside, I also want to meet his standards. ;-)

justinas-vd:
Okay, I will check that discard overflows behave as I imagine they should.

justinas-vd:
Fixed excessive discard checks, which makes tests run reasonably fast now.

Also added limit tests for discard.

swatanabe:
AMDG

On 02/26/2018 03:13 PM, Justinas V. Daugmaudis wrote:
> Fixed excessive discard checks, which makes tests run reasonably fast now.
> 
> Also added limit tests for discard.
> 

I haven't actually tried this, but I suspect the
test will fail if you multiply maxdiscard by 4 instead
of 2 in test_max_discard.  (You're testing
with 2 dimensions, which cancels out the multiply
by 2, so the truncation in the assignment
at qrng_base.hpp:178 won't trigger.)

Also, the reason I uses 7-bits instead of
8-bits when I suggested this is to make
sure that the w parameter is respected.

In Christ,
Steven Watanabe

justinas-vd:
Ok, now I see your point. Will fix the tests sometime tomorrow. As of now, I added additional checks for bit count overflows.

justinas-vd:
@swatanabe More tests – done. Moreover, the code in discard_vector had a genuine bug, I reckon. The test now expects an exception in case of overflows in discard.

justinas-vd:
1. Documentation is complete. But it would be good to review/correct wording. @NAThompson ?
2. Stress-tested the generators with various funny types and bit length combinations, e.g., char, 7u, etc. No undefined behavious, errors get properly reported. However, I would say that multiprecision lib should probably support unsigned short for msb and lsb. But this is not a limitation in the generators per se.
3. All obvious potential overflow-related bugs in qrng_base were fixed. In particular, there was a corner case in == operator, when eng.discard(maxdiscard) == other.seed(0), i.e., because of overflow the engine at the very end of the sequence would be equal to the engine at the very start of the sequence.
4. The code properly compiles and passes tests with -DBOOST_RANDOM_DOXYGEN. No invalid code in headers, even in documentation part.

doc/concepts.qbk
```diff
@@ -199,6 +199,76 @@ the generator, for example to re-run a test suite at a later time.]
...
 
+[section Quasi-Random Number Generator]
+
+A quasi-random number generator is a __UniformRandomNumberGenerator which
+provides a deterministic sequence of quasi-random numbers, based on some
+algorithm and internal state. [classref boost::random::niederreiter_base2_engine
+Niederreiter base 2] generator is an example of such a [qrng quasi-random
+number generator]. The "quasi" modifier is used to denote more clearly that the
+values produced by such a generator are neither random nor pseudo-random, but
+they form a low discrepancy sequence. The intuitive idea is that a low discrepancy
```
NAThompson: The "quasi" modifier is used to indicate that the stream of vectors delivered by the generator form a low-discrepancy sequence.

doc/concepts.qbk
```diff
@@ -199,6 +199,76 @@ the generator, for example to re-run a test suite at a later time.]
...
+sequence is more evenly distributed than a pseudo random sequence would be.
+For example, if we generate a low discrepancy sequence of 2D points on a square,
+this square would be covered more evenly, and the number of points falling to any
+part of the square would be proportional to the number of points in the whole square.
+Such sequences share some properties of  random variables and in certain applications
+such as the quasi-Monte Carlo method their lower discrepancy is an important advantage.
+
+[note Quasi-Monte Carlo method uses a low-discrepancy sequence such as the
```
NAThompson: "The quasi-Monte Carlo method uses . . "

doc/concepts.qbk
```diff
@@ -199,6 +199,76 @@ the generator, for example to re-run a test suite at a later time.]
...
+Niederreiter base 2 sequence, the Sobol sequence, or the Faure sequence among the others.
+The advantage of using low-discrepancy sequences is a probabilistically faster rate of
+convergence. Quasi-Monte Carlo has a rate of convergence close to '''<code>O(1/N)</code>''',
+whereas the rate of convergence for the Monte Carlo method, which uses a pseudo-random sequence,
+is '''<code>O(N<superscript>-0.5</superscript>)</code>'''.]
+
```
NAThompson: Quasi-Monte Carlo integration converges roughly as '''<code>O(1/N)</code>''', whereas the convergence of naive Monte Carlo integration, which uses pseudo-random sequences, is only '''<code>O(N<superscript>-0.5</superscript>)</code>'''.]

NAThompson:
Just a few typos fixed here. Where is the example code?

doc/concepts.qbk
```diff
@@ -199,6 +199,76 @@ the generator, for example to re-run a test suite at a later time.]
 
 [endsect]
 
+[section Quasi-Random Number Generator]
+
+A quasi-random number generator is a __UniformRandomNumberGenerator which
+provides a deterministic sequence of quasi-random numbers, based on some
```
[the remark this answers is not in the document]
justinas-vd: Sounds kinda oxymoronic, no? I would suggest just to change "numbers" to "values" so that we don't inadvertently imply that qrngs work in R^1 only.

doc/concepts.qbk
justinas-vd: Ok.

doc/concepts.qbk
justinas-vd: If we use "values" above, then this wording should be kept as it is. There is nothing wrong with it.

doc/concepts.qbk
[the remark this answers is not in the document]
NAThompson: Sorry, "a deterministic sequence of vectors"

doc/concepts.qbk
justinas-vd: No. The suggestion is less precise. Consider what "converges roughly as O(1/N)" is supposed to mean?

justinas-vd:
Doxygenized comments are also documentation. Mostly, what's important are comments before *_engine classes and before default specializations for those engines.

As for the example, what kind of example should I provide?

NAThompson:
I would just put should how the code is used and state the interface in the docs. See: 

http://www.boost.org/doc/libs/1_66_0/libs/math/doc/html/math_toolkit/gauss_kronrod.html

NAThompson:
Note that the background info on quasi-random sequences is not that important, since most people will navigate from the Wikipedia article on QRNGs to the boost docs, not the other way around.

justinas-vd:
Hm, I can write a very short tutorial how to use a 4-dimensional qrng to uniformly cover an S^2 sphere. I would like to keep the tutorial to the length that is typical of Boost.Random tutorials.

doc/concepts.qbk
NAThompson: Maybe return to the correct estimate of O(log(N)^d/N), as the logarithmic factors dominate for achievable N and somewhat large d.

NAThompson:
I think that would be excellent.

justinas-vd:
@NAThompson  Example – done.

NAThompson:
Looks good to me! For my own reference, could you also add a quick note about how to generate the entire vector at once? In your example, you extracted each element of the pseudo-random vector in different calls, which may be awkward in very high dimension.

justinas-vd:
This is what variate_generator is for. Plug-in the generator, the distribution and use std::generate(my_sample.begin(), my_sample.end(), gen());

swatanabe:
AMDG

On 03/06/2018 10:27 AM, Justinas V. Daugmaudis wrote:
> This is what variate_generator is for. Plug-in the generator, the distribution and use std::generate(my_sample.begin(), my_sample.end(), gen());
> 

variate_generator is obsolete.  Just use
std::bind(distribution, std::ref(engine))
or [&]() { return distribution(engine); }

In Christ,
Steven Watanabe

justinas-vd:
Thank you for the correction. This depreciation is news to me, but totally makes sense.

NAThompson:
It appears that the idiomatic method of generating the entire vector is not clear even to us, which is why I think it should be added to the docs.

It looks like everything is approaching its completed form. What are the final steps to complete this feature?

justinas-vd:
This mammoth of the PR still needs to be thoroughly reviewed by Steven. I did my best to write the generators that do exactly as advertised, and I am fairly confident about the algorithmic side of things, but we still need to verify this.

include/boost/random/detail/gray_coded_qrng.hpp
```diff
@@ -0,0 +1,152 @@
...
+  //!
+  //!Throws: nothing.
+  static BOOST_CONSTEXPR result_type min BOOST_PREVENT_MACRO_SUBSTITUTION ()
+  { return 0; }
+
+  //!Returns: Tight upper bound on the set of values returned by operator().
+  //!
+  //!Throws: nothing.
+  static BOOST_CONSTEXPR result_type max BOOST_PREVENT_MACRO_SUBSTITUTION ()
+  { return (std::numeric_limits<result_type>::max)(); }
```
swatanabe: This doesn't respect bit_count.  Use boost::low_bits_mask_t (from boost/integer/integer_mask.hpp).

include/boost/random/detail/qrng_base.hpp
```diff
@@ -0,0 +1,254 @@
...
+//!\file
+//!Describes the quasi-random number generator base class template.
+
+namespace boost {
+namespace random {
+
+namespace detail {
+
+template <typename T>
+inline bool are_close(T x, T y, T eps) // |x - y| <= eps
```
swatanabe: This function is unnecessary.  The only place you're using this is to compare integers.

include/boost/random/detail/qrng_base.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  //!Returns true if the two generators will produce identical sequences of outputs.
+  BOOST_RANDOM_DETAIL_EQUALITY_OPERATOR(qrng_base, x, y)
+  {
+    const std::size_t dimension_value = x.dimension();
+
+    // Note that two generators with different seq_counts and curr_elems can
+    // produce the same sequence because the generator triple
+    // (D, S, D) is equivalent to (D, S + 1, 0), where D is dimension, S -- seq_count,
+    // and the last one is curr_elem.
```
swatanabe: Would it make sense to normalize the representation?

include/boost/random/faure.hpp
```diff
@@ -0,0 +1,363 @@
...
+//!Volume 12, Number 4, December 1986, pages 362-376.
+//! \endblockquote
+//!
+//!In the following documentation @c X denotes the concrete class of the template
+//!faure_engine returning objects of type @c RealType, u and v are the values of @c X.
+//!
+//!Some member functions may throw exceptions of type @c std::bad_alloc.
+template<typename RealType, typename SeqSizeT, typename PrimeTable = default_faure_prime_table>
+class faure_engine
+  : public detail::qrng_base<
```
swatanabe: This has the unfortunate side-effect that boost::random::detail is an associated namespace of faure_engine.  I generally avoid putting internal base classes in namespace detail.  (You can also put qrng_base in its own namespace).

include/boost/random/detail/qrng_base.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+    // Avoid overdiscarding by branchlessly correcting the triple
+    // (D, S + 1, 0) to (D, S, D) (see equality operator)
+    const bool corr = (!carry) & static_cast<bool>(vec_n);
+
+    // Discards vec_n (with correction) consecutive s-dimensional vectors
+    discard_vector(vec_n - static_cast<boost::uintmax_t>(corr));
+    // Sets up the proper position of the element-to-read
+    // curr_elem = carry + corr*dimension_value
+    curr_elem = carry ^ (-static_cast<std::size_t>(corr) & dimension_value);
```
swatanabe: MSVC warns: C4146 unary minus operator applied to an unsigned type, result still unsigned.  (This should be safe to suppress).

include/boost/random/detail/qrng_base.hpp
```diff
@@ -0,0 +1,254 @@
...
+  void discard_vector(boost::uintmax_t z)
+  {
+    const size_type max_z = std::numeric_limits<size_type>::max() - seq_count;
+
+    // Don't allow seq_count + z overflows here
+    if (max_z < z)
+      boost::throw_exception( std::range_error("qrng_base: discard_vector") );
+
+    std::size_t tmp = curr_elem;
+    derived().seed(seq_count + z);
```
swatanabe: Add an explicit cast to suppress warnings here.

include/boost/random/detail/gray_coded_qrng.hpp
justinas-vd: Nice idea, but if I implement max() as 
{ return low_bits_mask_t<LatticeT::bit_count>::sig_bits; }

then 8 bit discard test doesn't compile because boost/integer/integer_mask.hpp:66:5: error: left operand of shift expression ‘(-1 << 7)’ is negative [-fpermissive]
     BOOST_STATIC_CONSTANT( least, sig_bits = (~( ~(least( 0u )) << Bits )) );

Basically, least is resolved to char, and it gets negative.

For the time being I would suggest to write this as (std::numeric_limits<result_type>::max)() >> (std::numeric_limits<IntType>::digits - LatticeT::bit_count).

This will produce the result we want until this problem with integer_mask gets resolved.

include/boost/random/detail/qrng_base.hpp
justinas-vd: Yes, it is. g++ doesn't complain here at all.

include/boost/random/detail/qrng_base.hpp
justinas-vd: It is normalized. The difference between (D, S, D) and (D, S + 1, 0) is that the former has all the elements exhausted but the new state is not yet computed (because not yet requested), while the latter is a freshly computed S + 1 state (because it was requested).

It is useful to maintain this difference, because otherwise if you do engine.discard(maxdiscard) and compute straight to state (D, last_state + 1, 0), you would be overdiscarding, you wouldn't have the equality between maxdiscard operator() invocations and discard(maxdiscard).

include/boost/random/detail/qrng_base.hpp
justinas-vd: I dislike this orphan function that is used once, too. But how would you suggest to compactly write the conjunctions in operator== without losing code flow?

swatanabe:
AMDG

On 03/06/2018 02:03 PM, Justinas V. Daugmaudis wrote:
> 
> +    // Note that two generators with different seq_counts and curr_elems can
> +    // produce the same sequence because the generator triple
> +    // (D, S, D) is equivalent to (D, S + 1, 0), where D is dimension, S -- seq_count,
> +    // and the last one is curr_elem.
> 
> It is normalized. The difference between (D, S, D) and (D, S + 1, 0) is that the former has all the elements exhausted but the new state is not yet computed (because not yet requested), while the latter is a freshly computed S + 1 state (because it was requested).
> 
> It is useful to maintain this difference, because otherwise if you do engine.discard(maxdiscard) and compute straight to state (D, last_state + 1, 0), you would be overdiscarding, you wouldn't have the equality between maxdiscard operator() invocations and discard(maxdiscard).
> 

  Is there ever a reason to want zero?  You're discarding
the zero vector, so setting the initial state to (D, 0, D)
should be fine.  operator() also never leaves curr_elem
as 0.

In Christ,
Steven Watanabe

justinas-vd:
Yes, set engine.seed(0) and you will have (D, 0, 0). Also, engine.seed(1) will give (D, 1, 0), etc.

swatanabe:
AMDG

On 03/06/2018 01:47 PM, Justinas V. Daugmaudis wrote:
> 
> then 8 bit discard test doesn't compile because boost/integer/integer_mask.hpp:66:5: error: left operand of shift expression ‘(-1 << 7)’ is negative [-fpermissive]
>      BOOST_STATIC_CONSTANT( least, sig_bits = (~( ~(least( 0u )) << Bits )) );
> 

  This has been fixed in develop (by inserting casts).
least is actually unsigned char.  The problem is the
implicit promotion to int.

> Basically, least is resolved to char, and it gets negative.
> 

In Christ
Steven Watanabe

include/boost/random/faure.hpp
justinas-vd: I see, I see... Would it be acceptable to deepen the namespace to, say, detail::quasi::qrng_base?
IIRC, ADL would get into effect only for the innermost enclosing namespaces, no?

justinas-vd:
Yay, works and tests pass! Thanks for removing that max() computation ugliness.

include/boost/random/faure.hpp
justinas-vd: I think this is done now. I moved qrng_base, gray_coded_qrng from namespace detail to qrng_detail.

justinas-vd:
@swatanabe  Done. Please check.

justinas-vd:
@swatanabe  Did you have a chance to take a look? Personally, I think that things could be improved by ensuring strong exception safety in `niederreiter_base2` and `sobol` lattices. We are resizing them anyway, so it would make sense to simply compute into the new memory and swap at the end. What would you suggest?

swatanabe:
AMDG

On 03/13/2018 04:31 AM, Justinas V. Daugmaudis wrote:
> @swatanabe  Did you have a chance to take a look? Personally, I think that things could be improved by ensuring strong exception safety in `niederreiter_base2` and `sobol` lattices. We are resizing them anyway, so it would make sense to simply compute into the new memory and swap at the end. What would you suggest?
> 

resize is only called from the constructor (strong
exception safety is irrelevant) and the istream
operator (this is not exception safe and must be
fixed.  It can currently leave the lattice in
a state inconsistent with quasi_state, which can
potentially lead to out-of-bounds indexes).  Hmm.
Strong exception safety doesn't really cost anything
extra, so you might as well provide the strong guarantee.

In Christ,
Steven Watanabe

justinas-vd:
Done.

NAThompson:
Excellent work Justinas. Is this ready to be merged into develop? If so, that gives me the opportunity to work on the integrator and perhaps discover some bugs via usage before release into 1.68.

justinas-vd:
I am confident the code works; seed, discard functions have no hidden corner cases. Thanks to Steven test coverage is much better now. For fun and giggles I instantiated niederreiter_base2_engine with <char, 7u>. It produced a normal sequence, albeit a very short one (as could be expected).

At this point I don't know what improvements I could make. In the future, however, I would like to reuse `powm` from multiprecision, provided that `integer_modulus` could be disabled by passing, say, mpl::false_ instead of integral `c`. I would like to reuse the prime table from Boost.Math, if we can add functions not only to get the nth prime, but also a prime that would be a lower bound of `n` (upper bound of `n` would come for free at that point). Those would be useful functions in and of themselves.

_Summa summarum_: the final say is Steven's and I know perfectly well that expert programmers have an excellent sense of intuition if something not quite kosher goes on in the code. This intuition has to be satisfied. By adding more tests, if need be.

justinas-vd:
@swatanabe Any comments, further improvements, test suggestions?

NAThompson:
Now that 1.67 has been released, is there any chance this can be merged?

swatanabe:
AMDG

On 04/18/2018 08:09 AM, Nick wrote:
> Now that 1.67 has been released, is there any chance this can be merged?
> 

I'll try to get to it this week.

In Christ,
Steven Watanabe

justinas-vd:
@swatanabe Hi, Steven. Did you have a chance to take a look?

swatanabe:
AMDG

On 05/03/2018 04:58 AM, Justinas V. Daugmaudis wrote:
> @swatanabe Hi, Steven. Did you have a chance to take a look?
> 

I've started merging it.
pass: gcc-7.3.1, clang-5.0.1 (-std=c++03/11/14/17)
fail: vc9, vc10 (Not high priority, but
  it should only take minor tweaks to fix).
pass: vc11, vc12, vc14, vc141

In Christ,
Steven Watanabe

justinas-vd:
Oh, wonderful news! I assume that failures for vc9 and vc10 are pretty much the same?

swatanabe:
AMDG

On 05/07/2018 01:26 PM, Justinas V. Daugmaudis wrote:
> Oh, wonderful news! I assume that failures for vc9 and vc10 are pretty much the same?
> 

Right.  I can fix it by adding parentheses.

Additional changes
- bit_xor is not in c++03 (vc9, not worth fixing)
- suppress C4800 (int to bool conversion)
- niederreiter_base2.hpp:79: C4146 (unary minus on unsigned.
  Signed integer overflow is undefined behavior, so I rewrote
  the loop to go from 0 to pb_degree)
- faure.hpp: Add explicit casts to suppress C4244.
- test_max_seed at test_qrng_function.cpp:148 doesn't
  work for faure, because eng.max is 1.0 and has nothing
  to do with the maximum seed.  Oh, I see that even though
  it's instantiated for faure, it isn't actually used.
  I'll leave this as is.

In Christ,
Steven Watanabe

NAThompson:
Great work guys; both of you put in a huge effort into making this happen. I'll get wikipedia updated so that people know that this feature exists and hopefully it'll be useful.

justinas-vd:
Thank you, Steven and Nick. Good work, indeed. Moreover, should the public uncover some bug that is unknown to us a this moment, I will certainly fix it. (I have a vested interest to see this code correct. I use it for my own simulations.)

PR #41, opened 2018-08-09
random negative binomials with fractional size

madeleineth:
Support negative binomial random generation with a fractional size parameter. Since it is cast to `RealType` anyway when it is used, I believe this is backward compatible with nearly all use cases.

For more information on why you would want a fractional size parameter, see https://www.johndcook.com/blog/2009/09/22/negative-binomial-distribution/; `boost::math::negative_binomial` already does.

(This is my first pull request to Boost; please let me know if you would like things handled differently. Neither I nor my employer, Tableau, claim copyright on this patch.)

### DIFF
--- a/include/boost/random/negative_binomial_distribution.hpp
+++ b/include/boost/random/negative_binomial_distribution.hpp
@@ -48,11 +48,11 @@ public:
          *
          * Requires: k >=0 && 0 <= p <= 1
          */
-        explicit param_type(IntType k_arg = 1, RealType p_arg = RealType (0.5))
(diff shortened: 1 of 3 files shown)


swatanabe:
AMDG

On 08/08/2018 06:51 PM, Madeleine Thompson wrote:
> Support negative binomial random generation with a fractional size parameter. Since it is cast to `RealType` anyway when it is used, I believe this is backward compatible with nearly all use cases.
> 

I'm sure I've seen this before.  Ah, here's
the discussion:
https://svn.boost.org/trac10/ticket/10110

In Christ,
Steven Watanabe

PR #55, opened 2019-04-28
Add basic CMake build system

samcday:
Just directly ripping off the work I've noticed @Mike-Devel doing across a bunch of Boost modules.

I determined the list of dependencies by grepping for `#include` directives. I don't know this library very well, so let me know if any of the dependencies don't make sense.

This PR depends on:

 * [dynamic_bitset](https://github.com/boostorg/dynamic_bitset/pull/44)
 * [multiprecision](https://github.com/boostorg/multiprecision/pull/131) (cyclic!)
 * [range](https://github.com/boostorg/range/pull/92)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,37 @@
+# Copyright 2019 Sam Day
+# Distributed under the Boost Software License, Version 1.0.
+# See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
+#
+# NOTE: CMake support for Boost.random is currently experimental at best
+#       and the interface is likely to change in the future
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

On 4/28/19 9:47 AM, Sam wrote:
> Just directly ripping off the work I've noticed @Mike-Devel doing across a bunch of Boost modules.
> 
> I determined the list of dependencies by grepping for `#include` directives. I don't know this library very well, so let me know if any of the dependencies don't make sense.
> 

I have no interest in maintaining CMake build files
myself.  Therefore, if you are not confident that
this is correct, I'm not going to accept it.

In Christ,
Steven Watanabe

PR #64, opened 2020-02-26
Fix -Wtautological-overlap-compare clang warning in linear_congruenti…

Romain-Geissler-1A:
…al.hpp

Clang 10 shows this warning/error (with -Werror):

/remote/intdeliv/components/osp/Boost/19-0-0-18/include/boost/random/linear_congruential.hpp:140:20: error: overlapping comparisons always evaluate to false [-Werror,-Wtautological-overlap-compare]
        if(_x <= 0 && _x != 0) {
           ~~~~~~~~^~~~~~~~~~
/remote/intdeliv/components/osp/Boost/19-0-0-18/include/boost/random/linear_congruential.hpp:393:11: note: in instantiation of member function 'boost::random::linear_congruential_engine<unsigned long, 25214903917, 11, 281474976710656>::seed' requested here
    { lcf.seed(cnv(x0)); }

### DIFF
--- a/include/boost/random/linear_congruential.hpp
+++ b/include/boost/random/linear_congruential.hpp
@@ -137,7 +137,7 @@ public:
             _x = x0 % modulus;
         }
         // handle negative seeds
-        if(_x <= 0 && _x != 0) {
+        if(_x < 0) {
             _x += modulus;
         }
         // adjust to the correct range


swatanabe:
AMDG

This is quite unfortunate, as the code in question is written
the way it is to avoid a different warning.

In Christ,
Steven Watanabe

Romain-Geissler-1A:
So what do you suggest then ? What was this other warning ?

Shall we use a pragma to silence it then ?

swatanabe:
AMDG

On 4/8/20 2:58 AM, Romain Geissler @ Amadeus wrote:
> So what do you suggest then ? What was this other warning ?
> 

I don't remember.  It looks like it comes from

commit 6e35d9a732f37928d5fc45fa32a3c69e4889a9ad
Author: Steven Watanabe <steven@providere-consulting.com>
Date:   Sat Nov 14 21:07:31 2009 +0000

so the warning may be completely obsolete by now.

> Shall we use a pragma to silence it then ?
> 

If no current compiler (gcc/clang/msvc) warns on your
code with max warnings enabled, then your patch should
be fine.  If there is still an issue I'd prefer to
accept your patch and suppress the other warning, as
the code is cleaner that way.

In Christ,
Steven Watanabe

Romain-Geissler-1A:
I compile my projects with `-Wall`, `-Wextra`, `-Werror`, and both gcc (current trunk) and clang (current trunk as well). I have not tried released version. This code doesn't seem to raise any more warning for these two cases.

I am not working at all in a Windows environment, and I really don't know anything about msvc, so I can't test there.

Romain-Geissler-1A:
So, do you think we should merge this in the end ?

PR #65, opened 2020-03-29
clang-diagnostic-tautological-overlap-compare

crazyhappygame:
```
boost/random/linear_congruential.hpp:140:20: error: overlapping comparisons always evaluate to false [clang-diagnostic-tautological-overlap-compare,-warnings-as-errors]
        if(_x <= 0 && _x != 0) {
```

### DIFF
--- a/include/boost/random/linear_congruential.hpp
+++ b/include/boost/random/linear_congruential.hpp
@@ -137,7 +137,7 @@ public:
             _x = x0 % modulus;
         }
         // handle negative seeds
-        if(_x <= 0 && _x != 0) {
+        if(_x <= 0) {
             _x += modulus;
         }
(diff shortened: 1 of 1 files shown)


crazyhappygame:
Could you merge it?

include/boost/random/linear_congruential.hpp
```diff
@@ -137,7 +137,7 @@ class linear_congruential_engine
             _x = x0 % modulus;
         }
         // handle negative seeds
-        if(_x <= 0 && _x != 0) {
+        if(_x <= 0) {
```
Kojoley, line 140: I think you simplified it wrong, it should be `_x < 0` (like the comment above says). Also, there is already a PR #64 for this.

PR #70, opened 2020-08-23
Initial CMake build system configuration implementation.

nemothenoone:
This PR contains CMake build system implementation with:
* Support of potential future Boost.CMake-fication.
* Support of testing.
* Support of using Boost.Random as a out-of-Boost library.

What is still missing:
* Building examples via CMake. (coming soon, in this PR or in a separate one)
* Building extra via CMake. (coming soon, in this PR or in a separate one)
* Performance testing via CMake. (coming soon, in this PR or in a separate one)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,67 @@
+#---------------------------------------------------------------------------#
+# Copyright (c) 2018-2020 Mikhail Komarov <nemo@nil.foundation>
+#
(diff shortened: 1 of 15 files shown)


Flamefire:
This has since been done in 1e9bfb8b3232093646df4c491e01a4f9667be783 using the "standard" workflow of other Boost libs

PR #75, opened 2020-12-23
 add BOOST_PREVENT_MACRO_SUBSTITUTION() to min,max

kotika:
Fix https://github.com/boostorg/math/issues/468, which is in turn caused by the collision in minwindef.h on Windows.h


evanmiller:
My understanding is that `BOOST_PREVENT_MACRO_SUBSTITUTION` should be between `min` and `()`, i.e.

```
BOOST_STATIC_CONSTEXPR result_type min BOOST_PREVENT_MACRO_SUBSTITUTION () {return mixmax_min;}
```

Instead of

```
BOOST_STATIC_CONSTEXPR result_type min() BOOST_PREVENT_MACRO_SUBSTITUTION() {return mixmax_min;}
```

evanmiller:
Usage example

https://www.boost.org/doc/libs/1_75_0/boost/limits.hpp

kotika:
Superceded.

PR #78, opened 2021-01-01
Remove dependence on multiprecision::lsb, msb

pdimov:
This pull request removes Random's dependence on Multiprecision, breaking the dependency cycle. Cycles are a problem for package managers that ship a modular Boost.

By explicitly throwing `std::range_error` from the helper lsb/msb functions it also fixes test errors in sobol_validate and niederreiter_base2_validate introduced by Multiprecision's functions being recently changed to (correctly) throw `std::domain_error` (input out of range) instead of `std::range_error` (output out of range).

I note that these generators also throw `std::range_error` on invalid seeds, which should in theory be `std::domain_error`, but fixing that is outside the scope of this PR.

### DIFF
--- a/include/boost/random/detail/gray_coded_qrng.hpp
+++ b/include/boost/random/detail/gray_coded_qrng.hpp
@@ -11,7 +11,9 @@
 
 #include <boost/random/detail/qrng_base.hpp>
 
-#include <boost/multiprecision/integer.hpp> // lsb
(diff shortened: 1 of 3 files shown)


pdimov:
(For the Multiprecision `std::range_error` change, see https://github.com/boostorg/multiprecision/issues/257 and https://github.com/boostorg/multiprecision/pull/281.)

PR #87, opened 2022-05-22
Fix comments in uniform_real_distribution

paulharris:
The comments state: `require min <= max`
but the code states `BOOST_ASSERT(_min < _max);`


paulharris:
This has one extra tiny change, compared with #4

swatanabe (approved):

PR #89, opened 2022-11-27
Add Splitmix64

mborland:
Adds the Splitmix64 generator from Sebastiano Vigna as described [here](https://dl.acm.org/doi/10.1145/2714064.2660195). Will be used for the future implementation (generates the initial seed) of the [xoroshiro family of PRNGs](https://prng.di.unimi.it). The performance comparison of the STL and current Boost.Random random to Splitmix64 is shown below:

```
Running ./random_device
Run on (10 X 24.1211 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x10)
Load Average: 1.41, 1.62, 1.84
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
stl_random_device<unsigned>            499 ns          498 ns      1039378
boost_random_device<unsigned>          500 ns          500 ns      1399804
boost_splitmix64<std::uint64_t>       5.23 ns         5.23 ns    133605634
```


mborland:
@NAThompson I assume you will have some interest in this since you have asked for other PRNGs before.

NAThompson:
@mborland : std::random_device is a hardware TRNG, whereas splitmix64 is a PRNG right? Shouldn't the comparison be to (say) PCG32?

include/boost/random/splitmix64.hpp
```diff
@@ -0,0 +1,98 @@
+/* 
+ * Copyright Sebastiano Vigna 2015.
```
NAThompson, line 2: Did Sebastiano Vigna write this code?

performance/random_device.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+template <typename T>
+void boost_splitmix64(benchmark::State& state)
+{
+    boost::random::splitmix64 rd;
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(rd());
```
NAThompson, line 40: For PRNGs, it is often necessary to do multiple calls to get accurate performance numbers because the loop overhead is so high.

test/test_splitmix64.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <boost/random/splitmix64.hpp>
+
+int main(void)
+{
+    constexpr auto const_min = (std::numeric_limits<std::uint64_t>::min)();
+    constexpr auto const_max = (std::numeric_limits<std::uint64_t>::max)();
+    
+    boost::random::splitmix64 rng;
+    
+    double entropy = rng.entropy();
```
NAThompson, line 23: I confess to not understanding the `.entropy()` member here; is this the entropy of a uniform distribution?

mborland: Since this is a [PRNG](https://en.cppreference.com/w/cpp/numeric/random/random_device/entropy) it is always 0 so not overly useful. Removed.

include/boost/random/splitmix64.hpp
mborland: The implementation of `next` is almost identical to his published C code. His code is all licensed under the Creative Commons license which is compatible with the Boost Software License.

mborland:
> @mborland : std::random_device is a hardware TRNG, whereas splitmix64 is a PRNG right? Shouldn't the comparison be to (say) PCG32?

You are correct. Here is a more accurate comparison of 64-bit PRNGs:

```
Running ./random_device
Run on (10 X 24.0415 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x10)
Load Average: 1.72, 2.10, 2.47
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
stl_mt19937_64<std::uint64_t>          128 ns          128 ns      5110757
boost_mt19937_64<std::uint64_t>       72.0 ns         72.0 ns      9697574
boost_splitmix64<std::uint64_t>       38.2 ns         38.2 ns     18326430
```

NAThompson:
@mborland : For some reason these numbers seem unreasonably slow. I expect `mt19937_64` to execute in about 5 ns . . .

include/boost/random/splitmix64.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+class splitmix64
+{
+private:
+    std::uint64_t state_;
+
+public:
+    using result_type = std::uint64_t;
+    
+    splitmix64()
```
NAThompson, line 35: What about adding `splitmix64(int64_t seed = 0)` where `seed==0` means "choose a TRNG seed"?

test/test_splitmix64.cpp
```diff
@@ -0,0 +1,33 @@
...
+    for(int i = 0; i < 1000; ++i) 
+    {
+        std::uint64_t val = rng();
+        BOOST_TEST_GE(val, const_min);
+        BOOST_TEST_LE(val, const_max);
+    }
+
+    std::array<std::uint64_t, 1000> test_array {};
+    rng.generate(test_array.begin(), test_array.end());
+
```
NAThompson, line 31: Could you test that your implementation is compatible with various distributions? i.e., pipe it through std::normal_distribution and `std::uniform_distribution` and `std::gamma_distribution`?

It's very easy for an integer RNG to not meet the conceptual requirements to map into a continuous distribution.

include/boost/random/splitmix64.hpp
```diff
@@ -0,0 +1,87 @@
...
+        state_ = static_cast<std::uint64_t>(word1) << 32 | word2;
+    }
+
+    explicit splitmix64(std::uint64_t state) : state_ {state} {}
+    
+    // Copying is explicity deleted to avoid two generators with the same state by mistake
+    splitmix64(const splitmix64&) = delete;
+    splitmix64 operator=(const splitmix64&) = delete;
+
+    std::uint64_t next() noexcept
```
NAThompson, line 51: inline?

include/boost/random/splitmix64.hpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        std::uint64_t z {state_ += UINT64_C(0x9E3779B97F4A7C15)};
+	    z = (z ^ (z >> 30)) * UINT64_C(0xBF58476D1CE4E5B9);
+	    z = (z ^ (z >> 27)) * UINT64_C(0x94D049BB133111EB);
+	    
+        return z ^ (z >> 31);
+    }
+
+    std::uint64_t operator()() noexcept
```
NAThompson, line 60: inline?

include/boost/random/splitmix64.hpp
```diff
@@ -0,0 +1,87 @@
...
+        return (std::numeric_limits<std::uint64_t>::max)();
+    }
+
+    constexpr std::uint64_t (min)() const noexcept
+    {
+        return (std::numeric_limits<std::uint64_t>::min)();
+    }
+
+    template <typename FIter>
+    void generate(FIter first, FIter last)
```
NAThompson, line 76: Is a `generate` member standard in C++ PRNGs? Or will it be in C++2x?

NAThompson:
One unit test which might be useful: Generally for PRNGs, we have requirements like [this one](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine):


<h3><span class="mw-headline" id="Notes">Notes</span></h3>
<p>The Nth consecutive invocation of a default-constructed engine is required to produce the following value:
</p>


N | The random engine type | The value to produce
-- | -- | --
10000 | std::mt19937 | 4123659995
10000 | std::mt19937_64 | 9981545732273789042


Is there an analog for splitmix64? Of course, you could introduce your own definition and simply document it.

kotika:
> On Nov 27, 2022, at 23:00, Nick ***@***.***> wrote:
> @NAThompson commented on this pull request.
> 
> In include/boost/random/splitmix64.hpp <https://github.com/boostorg/random/pull/89#discussion_r1033008927>:
> 
> > @@ -0,0 +1,98 @@
> +/* 
> + * Copyright Sebastiano Vigna 2015.
> Did Sebastiano Vigna write this code?
> 


Alert:

1) you cannot safely use up more than a cubic root of the period of any generator. This generator's period is apparently less than 2^64 but its not stated how much exatly.
A cubic root of that is ~2^21. Thus you must not use this generator in any program that will run longer than a millisecond.
In particular, you would not use it to create large numbers of streams on a multiprocessor system.

2) This generator is not from Vigna, the authors are not known to be experts on rngs, and the principle of construction 
 is obscure. It is neither an LCG, nor MRG, nor LFSR, nor counter based crypto. What is it exactly? 

3) For industrial strength RNG on boost, use MIXMAX, by your truly. It is parallelizable and splittable:
boost::random::mixmax gen{seed1,seed2,seed3,seed4}; 
It is enough that at least one of the 32bit seeds is different by one bit. This gives you 2^128 guaranteed non-colliding streams that pass all tests.

Cheers,
Kostas

=========================================
Institute of Nuclear and Particle Physics
NCSR Demokritos
http://inspirehep.net/author/profile/K.G.Savvidy.1
https://mixmax.hepforge.org

mborland:
Kostas,

The primary use for this is to seed the xoshiro/xoroshiro family of generators as recommended by the authors [here](https://dl.acm.org/doi/10.1145/1276927.1276928). The original C code is copyrighted by Vigna and David Blackman for the generators. I think it would be disingenuous to say that are not experts given the [large amount of literature they have published on the subject.](https://vigna.di.unimi.it/papers.php) A [large number of languages](https://prng.di.unimi.it/#quality) use these as their standard PRNG so I think it would be useful if we offered them as well.

include/boost/random/splitmix64.hpp
mborland: Inline is implicit for in-class definition, but does not hurt to make it explicit.

include/boost/random/splitmix64.hpp
mborland: It is not a standard member in the STL, but the Boost versions have this functionality so for the sake of consistency I added it.

swatanabe:
On 11/28/22 14:56, Matt Borland wrote:
> @mborland commented on this pull request.
> 
> +    template <typename FIter>
> +    void generate(FIter first, FIter last)
> 
> It is not a standard member in the STL, but the Boost versions have this functionality so for the sake of consistency I added it.
> 

The reason it exists is to make every generator usable as a SeedSeq,
for compatibility with ancient (pre-C++11) Boost.Random.  It's not
very relevant anymore.

swatanabe:
On 11/27/22 20:53, Nick wrote:
> Could you test that your implementation is compatible with various distributions? i.e., pipe it through std::normal_distribution and `std::uniform_distribution` and `std::gamma_distribution`?
> 
> It's very easy for an integer RNG to not meet the conceptual requirements to map into a continuous distribution.
> 

Please do use the generic test harness for random engines. It has some
obsolete functionality, like the iterator seed routines and the generate
function that should probably not be supported for new engines and
should be #ifdef'ed out in test_generator.ipp, but it will catch
things like the missing seed_seq constructor and the missing
seed function.

Also, the implementation of generate is wrong. generate is supposed
fill a range with 32-bit values, regardless of the both the engine's
value_type and the array's element type.

kotika:
> On Nov 28, 2022, at 23:53, Matt Borland ***@***.***> wrote:
> 
> 
> Kostas,
> 
> The primary use for this is to seed the xoshiro/xoroshiro family of generators as recommended by the authors here <https://dl.acm.org/doi/10.1145/1276927.1276928>. The original C code is copyrighted by Vigna and David Blackman for the generators. I think it would be disingenuous to say that are not experts given the large amount of literature they have published on the subject. <https://vigna.di.unimi.it/papers.php> A large number of languages <https://prng.di.unimi.it/#quality> use these as their standard PRNG so I think it would be useful if we offered them as well.
> 



I am in favor of including the generator of Vigna, but the fact that some generator needs another "good" generator to seed it is evidence that
it has long internal decorrelation time, like the Mersenne Twister. And in fact Vigna's generators do have this unfortunate property. 
Ideally any new generators should be constructed with short decorrelation times
and seeded by skipping along, like my generator. It needs neither non-linear seeding, nor an output scrambling function.

At a practical level, why is this being done ahead of or instead of including an implementation of Vignas generator?

Cheers,
Kostas

=========================================
Institute of Nuclear and Particle Physics
NCSR Demokritos
http://inspirehep.net/author/profile/K.G.Savvidy.1
https://mixmax.hepforge.org

include/boost/random/splitmix64.hpp
jzmaddock: Just double checking this as creative commons isn't always BSL compatible, but there's an explicitly "public domain" version so this really is OK ;)

jzmaddock:
>The primary use for this is to seed the xoshiro/xoroshiro family of generators as recommended by the authors [here](https://dl.acm.org/doi/10.1145/1276927.1276928). The original C code is copyrighted by Vigna and David Blackman for the generators. I think it would be disingenuous to say that are not experts given the [large amount of literature they have published on the subject.](https://vigna.di.unimi.it/papers.php) A [large number of languages](https://prng.di.unimi.it/#quality) use these as their standard PRNG so I think it would be useful if we offered them as well.

+1, it's Java's default PRNG which is good enough reason to offer it as an option at least.  It's drawbacks should be well documented though.

mborland:
> On 11/27/22 20:53, Nick wrote: Could you test that your implementation is compatible with various distributions? i.e., pipe it through std::normal_distribution and `std::uniform_distribution` and `std::gamma_distribution`? It's very easy for an integer RNG to not meet the conceptual requirements to map into a continuous distribution.
> Please do use the generic test harness for random engines. It has some obsolete functionality, like the iterator seed routines and the generate function that should probably not be supported for new engines and should be #ifdef'ed out in test_generator.ipp, but it will catch things like the missing seed_seq constructor and the missing seed function. Also, the implementation of generate is wrong. generate is supposed fill a range with 32-bit values, regardless of the both the engine's value_type and the array's element type.

The generic test harness is used and now passes on CI. I removed two of the obsolete constructor tests you mentioned. I ended up leaving the generate function in otherwise I would have to remove all of the concepts tests. They seem valuable to retain.

mborland:
> On Nov 28, 2022, at 23:53, Matt Borland ***@***.***> wrote: Kostas, The primary use for this is to seed the xoshiro/xoroshiro family of generators as recommended by the authors here <https://dl.acm.org/doi/10.1145/1276927.1276928>. The original C code is copyrighted by Vigna and David Blackman for the generators. I think it would be disingenuous to say that are not experts given the large amount of literature they have published on the subject. <https://vigna.di.unimi.it/papers.php> A large number of languages <https://prng.di.unimi.it/#quality> use these as their standard PRNG so I think it would be useful if we offered them as well.
> I am in favor of including the generator of Vigna, but the fact that some generator needs another "good" generator to seed it is evidence that it has long internal decorrelation time, like the Mersenne Twister. And in fact Vigna's generators do have this unfortunate property. Ideally any new generators should be constructed with short decorrelation times and seeded by skipping along, like my generator. It needs neither non-linear seeding, nor an output scrambling function. At a practical level, why is this being done ahead of or instead of including an implementation of Vignas generator? Cheers, Kostas ========================================= Institute of Nuclear and Particle Physics NCSR Demokritos http://inspirehep.net/author/profile/K.G.Savvidy.1 https://mixmax.hepforge.org

I separated this from the implementation of the other generators for ease of review. This is over 300 lines, and to include it as an implementation detail of the generators seemed unwieldy to me.

kotika:
I dont understand, Vigna writes code based on a paper by Steele, puts his copyright on --  justly so  because he did not copy any code -- then you copy-paste and wrap it in boost, and put your own copyright onto it. Is this even kosher? The file has a Copyright at the top, even if lower down it has a Creative commons license. The license I suppose gives you the legal right to do so, but ....
 shouldn't it at least say the algorithm is by Steele et al? Some of the many versions of splitmix64 floating around the internet have 

> Fast Splittable Pseudorandom Number Generators
> Steele Jr, Guy L., Doug Lea, and Christine H. Flood. "Fast splittable
> pseudorandom number generators."
> ACM SIGPLAN Notices 49.10 (2014): 453-472.

mborland:
@kotika There is already a link to the paper on ACM from Steele et. al. in the comments at the top of the file.

@jzmaddock, @swatanabe, and @NAThompson I think this is good for review. This passes CI using the existing generic test harness and concept tests with tailoring for C++11 as recommended by Steve.

NAThompson:
@swatanabe : @mborland is now in the employ of boost. Would you mind giving him maintainer access in order that he could merge a few of his MRs?

jzmaddock:
Actually @swatanabe agreed to this off list, sorry been busy with other things, just give Matt write access now.

PR #90, opened 2022-12-01
Fix -Wtautological-overlap-compare

mborland:
Issue can be replicated using Clang 14 with default warnings.


mborland:
@swatanabe This should be good to merge.

Issue #112, opened 2024-10-23
Inconsistent runtime assertion usage of BOOST_ASSERT and assert


ygeunkim:
I noticed inconsistent usage of assertion in this random library.
While most distributions  use `BOOST_ASSERT` or static assert, some distributions use the standard `assert`:

- [beta_distribution.hpp](https://github.com/boostorg/random/blob/dcaee178a68566bbe50c5454a8ff3db67412055e/include/boost/random/beta_distribution.hpp#L51)
- [hyperexponential_distribution.hpp](https://github.com/boostorg/random/blob/dcaee178a68566bbe50c5454a8ff3db67412055e/include/boost/random/hyperexponential_distribution.hpp#L269)

Is there any specific reason to use `assert` here?

FYI, I'm using `BOOST_ENABLE_ASSERT_HANDLER` to customize `BOOST_ASSERT` behavior when using this library.
I cannot control the standard assert calls in the same way.

Would it be possible to standardize `BOOST_ASSERT` for consistency and better assertion control of users?

mborland:
> I noticed inconsistent usage of assertion in this random library. While most distributions use `BOOST_ASSERT` or static assert, some distributions use the standard `assert`:
> 
> * [beta_distribution.hpp](https://github.com/boostorg/random/blob/dcaee178a68566bbe50c5454a8ff3db67412055e/include/boost/random/beta_distribution.hpp#L51)
> * [hyperexponential_distribution.hpp](https://github.com/boostorg/random/blob/dcaee178a68566bbe50c5454a8ff3db67412055e/include/boost/random/hyperexponential_distribution.hpp#L269)
> 
> Is there any specific reason to use `assert` here?
> 

No. My guess is it was written by a different author and nobody had brought it up in the last decade since it was merged.

> FYI, I'm using `BOOST_ENABLE_ASSERT_HANDLER` to customize `BOOST_ASSERT` behavior when using this library. I cannot control the standard assert calls in the same way.
> 
> Would it be possible to standardize `BOOST_ASSERT` for consistency and better assertion control of users?

If you submit a PR I will take a look and merge it. Otherwise I'll make one in a few days.

PR #113, opened 2024-10-23
BOOST_ASSERT usage #112

ygeunkim:
- Use BOOST_ASSERT instead of a standard assert
- Leave `assert` in test code in this stage
- Related to #112

### DIFF
--- a/include/boost/random/beta_distribution.hpp
+++ b/include/boost/random/beta_distribution.hpp
@@ -48,8 +48,8 @@ public:
                             RealType beta_arg = RealType(1.0))
           : _alpha(alpha_arg), _beta(beta_arg)
         {
-            assert(alpha_arg > 0);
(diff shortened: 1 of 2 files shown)


mborland:
Thanks @ygeunkim. Let me know when you think this is good to go since it's marked as draft. I'll run the CI and review it.

ygeunkim:
Thanks @mborland! I've changed `assert` to `BOOST_ASSERT` in beta_distribution.hpp and hyperexponential_distribution.hpp mentioned in the issue #112, while didn't touch the standard assert in test codes just in case.

mborland (changes requested):
I would add `#include <boost/assert.hpp>` to both of these. Even if the header is transitively pulled it it's better to be explicit.

mborland (approved):

PR #115, opened 2024-11-08
feat: add option to disable `random_device` for header-only build

zjyhjqs:
As demonstrated in [documentation](http://www.boost.org/doc/libs/release/more/getting_started/windows.html#header-only-libraries), the random lib can be header-only if `random_device` is disabled.


zjyhjqs:
Please review the PR and merge it if possible.

mborland:
> Please review the PR and merge it if possible.

I approved the CI run. The change looks fine and is inline with what's already available when building with B2.

Issue #125, opened 2025-02-01
Adding new probability distributions?


ygeunkim:
Hi, I wonder if it is okay to add new probability distributions. I recently submitted a PR without opening an issue first. I think some discussions can help.
I'm considering two (one in PR and one planning) that are frequently used in Bayesian literature:

- Inverse Gaussian distribution in (#124) : Its density function exists in [math](https://www.boost.org/doc/libs/1_87_0/libs/math/doc/html/math_toolkit/dist_ref/dists/inverse_gaussian_dist.html) module, but random module does not provide its generator. IMHO, it would be great if there is corresponding generator. Schucany et al. (1976)'s algorithm is straightforward. It is also in Wikipedia.
- Generalized inverse Gaussian (GIG) distribution: This is widely used in MCMC. Several sampling algorithms exist, but Hormann and Leydold (2014) is the most widely adopted. The authors have already published an R package with C code: [GIGrvg](https://github.com/unuran/GIGrvg/blob/main/GIGrvg/src/GIGrvg.c). Since its parameter region division is somewhat different from the original paper, so citing both would be appropriate.

If maintainers are open to add these distributions, I'll submit additional PR for the GIG.
Looking forward to your thoughts!

mborland:
Thank you, and yes we are open to adding the distributions. I will review the PR on Monday. I took a quick look at linked C code repo. Since it is licensed under the GPL you cannot use that code or a modified version of it since the license is incompatible with the BSL-1.0 license.

ygeunkim:
Thank you for the clarification.
I'll use a different algorithm for the GIG to ensure compatibility and submit a PR accordingly.

ygeunkim:
Both PRs (#124 and #126 ) are merged.

PR #126, opened 2025-02-02
Feature: Generalized inverse Gaussian random generator

ygeunkim:
From Issue #125, added generalized inverse Gaussian distribution.

- Added `random/generalized_inverse_gaussian_distribution.hpp`
- Added `test/test_generalized_inverse_gaussian_distribution.cpp`
- Algorithm used: [Devroye (2014)](https://link.springer.com/article/10.1007/s11222-012-9367-z)

Need to check

- [ ] Naming: Name is quite long.
- [ ] Edge case: Whether to consider very small parameter case (can be covered using `gamma_distribution<>`)


ygeunkim:
Should replace std::cosh and std::sinh in old C++ version.

mborland (commented):
LGTM. Do you still need to think about the two items listed as need to check in your original PR comment? Is it possible to cook up some additional tests that will cover your two uncovered lines?

ygeunkim:
The algorithm has three cases but the test only covers up to 2 cases.
https://github.com/boostorg/random/blob/8378063ac7ff99645a7fb51a4b985e7ec43450a1/test/test_distribution.ipp#L209
I'll add it after adding BOOST_RANDOM_TEST3_* in test_distribution.ipp.

ygeunkim:
It seems there are still missing lines. Will change the numbers for them ASAP.

ygeunkim:
I checked codecov result, and it looks like every line is covered. Please review the changes. Since I updated the `test_distribution.ipp`, the merge might require a careful look.

mborland:
So this is what I see from the codecov report this morning. Since these are control paths for `operator()` we should have tests for them. If you need instead of expanding the `test_distribution.ipp` file you can add your own cpp file in tests directory. I find `<boost/core/lightweight_test.hpp>` easier to use: https://www.boost.org/doc/libs/master/libs/core/doc/html/core/lightweight_test.html

Screenshot 2025-02-05 at 9 52 11 AM

ygeunkim:
I should have added TEST3 in the test file. I mistakenly used TEST2 twice.
Also, I'll try your suggestion. Since my commit has added some lines in `test_distribution.ipp`, it might be better to revert it and try the lightweight_test.

ygeunkim:
Every line is now covered.

image

PR #135, opened 2025-06-30
Improve efficiency of `beta_distribution::operator()`

mborland:
@Hailios since `gamma_distribution::operator()` can be marked `const` we can in fact avoid re-initializing at least 2 gamma distributions for each call to `beta_distribution::operator()`.


mborland:
The static has been removed, and all diffs are on master so you'll see them in the 1.89 release in the next few weeks. Thanks again.

PR #156, opened 2026-09-05
Fix stream ops for RNG instances

Flamefire:
Users might want to test if reading or writing the RNG state worked
The only way is checking the stream state.

So add assertions for that and fix the offending ones


mborland (approved):
This looks good, but currently has a merge conflict.

Flamefire:
Rebased. CI passes again

PR #158, opened 2026-09-06
Add most tests to CMake build and fix some test dependencies

Flamefire:
Although BoostTestJamfile is used many tests are not included in the CMake build because they are not of the form `run source.cpp ;` but specify extra dependencies

I replaced that call by a small Jamfile parser that knows about the dependencies.

This surfaced a few dependency issues in the Jamefile:
- The dependency on Boost.Random is already set for all targets, so does not need to be added again
- boost_bind was missing from the math tests although given the C++11 base it could now be eliminated
- boost_iterator was missing from one test
- `test_random_device` was always compiled as static and shared, also compiling all dependencies twice. CI uses `./b2 link=static,shared` explicitly which I'd say is a cleaner way to test than basically forcing this

Also a mini-change to avoid the deprecation warning from Boost.Bind


mborland (approved):
This looks like a lot of overlap with #158, but since that one already has a conflict, I will merge this one.

Flamefire:
> This looks like a lot of overlap with #158, but since that one already has a conflict, I will merge this one.

You meant #156? I included this PR there to be able to debug that. Will rebase

PR #160, opened 2026-09-09
Fix multiprecision_float_test.cpp

Flamefire:
The isfinite check needs to be ran on each element. Introduce helper functions to abstract generic parts.


Flamefire:
Merged as part of #157

PR #162, opened 2026-09-10
Allow using unsecure Node version for CI

Flamefire:
Workaround until https://github.com/boostorg/random/pull/157 can be merged


.github/workflows/ci.yml
```diff
@@ -21,6 +21,7 @@ env:
   GIT_FETCH_JOBS: 8
   NET_RETRY_COUNT: 5
   DEFAULT_BUILD_VARIANT: debug,release
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
mborland, line 24: This will not work anymore in less than 2 weeks: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

mborland:
This repo doesn't have Drone setup for it yet so maybe that's the better option at this point? Drop everything that does not support actions v6 from GitHub Actions and add them to Drone instead.

.github/workflows/ci.yml
Flamefire: That's why I have #157 already done. However that was having a few issues and the latest CI builds with the current config are expired. So wanted to have at least one "original" run to compare although most issues seem to be only discovered by better test coverage (new compilers, code coverage on Windows, sanitizers)
Due to the failing CI my other PRs cannot be tested either.

So this 1-line change is just meant to buy some time for that.

Feel free to close this and merge the other one which I'd prefer.

Flamefire:
Drone can be added additionally for redundancy. The reusable workflow from Boost.CI is intended to keep GHA running and updated without each repo having to do the same work.
And I do like the status display from GHA much more with each check shown individually.

Quick check: About half (15/33) of the GHA CI checks would need to be removed or updated and all the used actions too

Anyway: Why not merge this now?
Currently CI does not work.
With this CI works for 2 weeks more and that line can be removed again when you get to #157 or switch to drone if you prefer that

I just added the workaround used by Boost.CI in #163 if you prefer that as an (intermediate) solution

Flamefire:
Closing in favor of #163 for a workaround. But I suggest using #157

PR #163, opened 2026-09-11
Update GHA actions and Node version

Flamefire:
Alternative to https://github.com/boostorg/random/pull/162 which works at least longer.

Better to switch to Boost.CI though, see #157


mborland:
Superseded by #157

PR #164, opened 2026-09-12
Cleanup superflous Boost headers

Flamefire:
In most cases `std::uint32_t` etc and in all cases `std::numeric_limits` is used.

So include the standard headers and replace the remaining Boost type usages

### DIFF
--- a/extra/haertel.hpp
+++ b/extra/haertel.hpp
@@ -18,7 +18,7 @@
 #ifndef BOOST_RANDOM_HAERTEL_HPP
 #define BOOST_RANDOM_HAERTEL_HPP
 
-#include <boost/cstdint.hpp>
+#include <cstdint>
 #include <boost/random/linear_congruential.hpp>
 #include <boost/random/inversive_congruential.hpp>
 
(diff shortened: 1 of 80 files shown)


mborland (approved):
Thanks! I need to look more closely at the rest of the dependencies in this lib. I'm sure almost all can go away.

Flamefire:
Given the amount of C++11 features already required I guess that allows dropping more, e.g. Boost.bind

Note that this library is still very valuable: C++11 does not define how distributions work (e.g. how many calls to the RNG occur) and how distributions and generators are serialized.
That's a big advantage of this implementation: It can be used where reproducible portability is required.

mborland:
> Given the amount of C++11 features already required I guess that allows dropping more, e.g. Boost.bind
> 

One of the dependencies is Math so we're at least up to C++14 here.

> Note that this library is still very valuable: C++11 does not define how distributions work (e.g. how many calls to the RNG occur) and how distributions and generators are serialized. That's a big advantage of this implementation: It can be used where reproducible portability is required.

For sure. We also have newer components with better properties than those offered by `<random>`

PR #166, opened 2026-09-13
Fix deprecated-copy warnings in polynomial.hpp

Flamefire:
Explicitly define the copy/move constructor as the implicit copy constructor for 'reference' is deprecated because it has a user-declared copy assignment operator

I moved `reference &operator^=(bool b)` to the bottom to avoid it being inside the conversion operators, so they can be closer together making reasoning about them easier

I noticed that `operator bool` should likely be explicit

Alternative to #111 after #110 got closed even though considered correct by @mborland


mborland (approved):
Looks good. Thanks! Will merge once the CI finishes.

PR #167, opened 2026-09-13
Remove Boost.Assign & Boost.Bind dependencies and switch to `BOOST_TEST`

Flamefire:
Replace Boost.Assign by initializer list and simple helper due to commas otherwise interfering with the macro
Mostly done by Regex-search&replace

This discovered undeclared dependencies on Boost.range

Single file used Boost.Bind -> Also replaced by `std::bind` already used elsewhere

While at it another S&R to use `BOOST_TEST` instead of `BOOST_CHECK<OP>`


mborland (approved):
Looks good. Thank you!

